Issue #13, opened 2021-05-05
[UI]: Dashboard: Started application card for multiple states


kcullion:
**Goal:** To set up the card for a started application, and have multiple states ready with the correct static data. 

# Expected Outcome
Have the following application cards set up with dummy data:
`Draft`, `Sign&Submit`, `DACO Review`, `Approved`, `Revisions Needed`, `Rejected`, `Closed` (before approval), `Closed` (after approval).

**Global card elements**
- [ ] 1. Application number
- [ ] 2. Institution name. If the applicant hasn't provided this, this can say` Institution: to be specified`
- [ ] 3. Progress bar as built in https://github.com/icgc-argo/dac-ui/issues/12
- [ ] 4. Last updated date

![image](https://user-images.githubusercontent.com/18424138/117222428-5fb2e480-add9-11eb-8a24-72da8a46aeaa.png)

**Custom Elements**

- [ ] 5. Status information sentence - this is different for each card. Do we want to hard code this for each state (but then the date is dynamic)?
- [ ] 6.`Approved` card has an Expiry date on the top right. Zeplin: https://zpl.io/2ZMEE7J

kcullion:
@rosibaj @ciaranschutte 
This looks good for now and we can test further once data is hooked up. I made this ticket and started the feedback - and moved the custom elements to test to the bottom of this ticket: https://github.com/icgc-argo/dac-ui/issues/122

Issue #12, opened 2021-05-05
[UI]: Application state progress bars


kcullion:
**Goal:** To set up the application progress bar component in storybook.
There's a similar component for ARGO to utilize: https://argo-ui-storybook.netlify.app/?path=/story/uikit-progress--progress-bar

These will be added to the dashboard card and the application page titlebar.

# Expected Outcome
Have the following progress bars set up for the following application states
- [ ] Draft: https://zpl.io/a8ngJGJ
![image](https://user-images.githubusercontent.com/18424138/117209968-16f13080-adc5-11eb-9cc5-ebf75161e915.png)
- [ ] Submit & Sign: https://zpl.io/29mQEYy
![image](https://user-images.githubusercontent.com/18424138/117210059-338d6880-adc5-11eb-8264-971af38668d7.png)
- [ ] DACO Review: https://zpl.io/bo3qGrZ
![image](https://user-images.githubusercontent.com/18424138/117210148-4c961980-adc5-11eb-93b0-b14f95c22053.png)
- [ ] Approved: https://zpl.io/2ZMEE7J
![image](https://user-images.githubusercontent.com/18424138/117210213-620b4380-adc5-11eb-9923-f0b4286b1915.png)
- [ ] Revisions Needed: https://zpl.io/Vx3zv5R
![image](https://user-images.githubusercontent.com/18424138/117210274-751e1380-adc5-11eb-8762-a08f6f637888.png)
- [ ] Rejected: https://zpl.io/VKdRLY6
![image](https://user-images.githubusercontent.com/18424138/117210330-8535f300-adc5-11eb-8b7f-477b28847c62.png)
- [ ] Closed (before approving): https://zpl.io/bJDmm8D
![image](https://user-images.githubusercontent.com/18424138/117210718-0f7e5700-adc6-11eb-947b-b0ac44864485.png)
- [ ] Closed (after approving): https://zpl.io/2j366Kx
![image](https://user-images.githubusercontent.com/18424138/117211015-7b60bf80-adc6-11eb-908d-1d203bb51543.png)

rosibaj:
@kcullion i am going to move this ticket forward. if ui feedback is needed we can incorporate that once data has been connected and we can see the states. 

Thank you @ciaranschutte !

daniel-cy-lu:
I didn't edit anything. I clicked and un-clicked one of the check-boxes by accident when I was studying the history on this epic.

caravinci:
[...]

PR #31, opened 2021-05-17
14 - add application page with basic titlebar

caravinci:
Creates the template for this:

image

in order to see it use `/applications/1234` or whatever fake ID you want to try.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/index.tsx
@@ -1,0 +1,214 @@
+import { ReactElement } from 'react';
+import { css } from '@icgc-argo/uikit';
+import Button from '@icgc-argo/uikit/Button';
+import Icon from '@icgc-argo/uikit/Icon';
+import { UikitTheme } from '@icgc-argo/uikit/index';
+import { ContentBody, ContentBox } from '@icgc-argo/uikit/PageLayout';
+import { useTheme } from '@icgc-argo/uikit/ThemeProvider';
+import Typography from '@icgc-argo/uikit/Typography';
+import VerticalTabs from '@icgc-argo/uikit/VerticalTabs';
+
+const ApplicationFormsBase = (): ReactElement => {
+  const theme: UikitTheme = useTheme();
+
+  return (
+    <ContentBody>
+      <ContentBox 
+        css={css`
+          box-sizing: border-box;
+          display: flex;
+          margin: 10px auto 0;
+          max-width: 1200px;
+          min-width: 665px;
+          width: 100%;
+        `}
+      >
+        <VerticalTabs
+          css={css`
+            background: ${theme.colors.grey_4};
+            border-radius: 8px 0 0 8px;
+            margin: -8px 0;
+            margin-left: -8px;
+            min-width: 205px;
+            max-width: 280px;
+          `}
+        >
+          <VerticalTabs.Item>
+            Table of Contents
+          </VerticalTabs.Item>
+          
+          <VerticalTabs.Item
+            active
+            onClick={function noRefCheck(){}}
+          >
+            Introduction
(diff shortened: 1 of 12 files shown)


components/pages/Applications/ApplicationForm/Header/Progress.tsx
```diff
@@ -0,0 +1,7 @@
+import { ReactElement } from "react";
+
+const ApplicationProgress = (): ReactElement => (
+  <section>progress bar here</section>
```
caravinci, line 4: Placeholder for the progress indicator to be added in #12 by @ciaranschutte

components/pages/Applications/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }: QueryType = useRouter();
+  
+
+  return (
+    <DefaultPageLayout title={'Application page'} >
+      {applicationID ? (
+        <ApplicationForm ID={applicationID} />
+      ) : ( // placeholder for the Dashboard
+        <> 
+        </>
```
caravinci, line 25: will contain #13 etc by @ciaranschutte

package.json
```diff
@@ -11,7 +11,7 @@
     "@emotion/babel-plugin": "^11.3.0",
     "@emotion/core": "^10.0.10",
     "@emotion/styled": "^10.0.11",
-    "@icgc-argo/uikit": "^1.9.0",
+    "@icgc-argo/uikit": "^1.9.1",
```
caravinci, line 14: adds purple tags and form icons

pages/applications/[[...ID]].tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactElement } from 'react';
+
+import Applications from 'components/pages/Applications';
+
+const ApplicationPage = (): ReactElement => {
+  return <Applications />;
+};
+
+export default ApplicationPage;
```
Comment (caravinci) [inline pages/applications/[[...ID]].tsx:9]: this filename takes url params to be passed into the page itself

components/pages/Applications/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type QueryType = {
+  query: {
+    ID?: string[];
+  };
+};
+
+const Application = (): ReactElement => {
+  const {
+    query: { ID: [applicationID] = [] },
```
caravinci, line 15: gets the ID from the URL, as given by NextJS

caravinci (commented):
A few comments to assist reviewing

pages/applications/[[...ID]].tsx
Comment (ciaranschutte) [reply pages/applications/[[...ID]].tsx:9]: `...whatever` is also catch all so any additional  `applications/12345/random-gibberish` will match too
we can remove the catch all part if we need too :)

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+            onClick={function noRefCheck(){}}
+          >
+            Sign &amp; Submit
+            {/* <VerticalTabs.Tag variant="SUCCESS">
+              45
+            </VerticalTabs.Tag> */}
+          </VerticalTabs.Item>
+        </VerticalTabs>
+
+        <main
```
ciaranschutte, line 132: I think we should have one `main` html element. We already have one our DefaultPageLayout

ciaranschutte (approved):
Looks great! Only feedback would be to have one `main` element

components/pages/Applications/ApplicationForm/Forms/index.tsx
caravinci: Agreed. After doing some reading, it seems the usage of `main` has changed since when they first introduced html5. 
Back then, iirc, different articles could have their own headers, footers, mains and asides; but now the suggestion is indeed one main per document. 

Will amend this before merging!

PR #33, opened 2021-05-18
fix build issues

caravinci:


components/DefaultPageLayout.tsx
```diff
@@ -11,7 +11,7 @@ const ThreeRowPageContainer = styled(PageContainer)`
   grid-template-rows: 58px 1fr 110px;
 `;
 
-const DefaultPageLayout = ({ children, title = '' }: { children: ReactElement, title?: string }): ReactElement => {
+const DefaultPageLayout = ({ children, title = '' }: { children: ReactElement | ReactNode, title?: string }): ReactElement => {
```
ciaranschutte, line 14: ReactNode I think covers ReactElement? (superset?)

caravinci: tried that earlier, but didn't seem to cover it before. weird 🤔

components/pages/Home/index.tsx
```diff
@@ -3,7 +3,11 @@ import React from 'react';
 import DefaultPageLayout from 'components/DefaultPageLayout';
 
 const Home = () => {
-  return <DefaultPageLayout title={'Homepage'} />;
+  return (
+    <DefaultPageLayout title={'Homepage'}>
+      &nbsp;
+    </DefaultPageLayout>
```
ciaranschutte, line 9: guessing this is just a prettier thing :)

ciaranschutte (approved):

components/pages/Home/index.tsx
caravinci: no, the `children` props was required, so I added a placeholder for the home page... but instead made the children optional.

PR #42, opened 2021-05-21
change download pdf button label

caravinci:
Just changing the label of the button, plus linting noise.


components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -39,10 +31,10 @@ const HeaderActions = (): ReactElement => {
           height="12px"
           name="download"
         />{' '}
-        Download PDF
+        Draft PDF
```
caravinci, line 34: this here is the gold nugget

ciaranschutte (approved):

PR #46, opened 2021-05-25
✨ 3 - login and routes

anncatton:
Adds Auth provider, with login and logout functionality. Using localStorage for ego jwt
Adds egoTokenUtils for permissions. Permissions available in Auth context, but not used yet
Adds config values for api, ego and external links
Updates navbar to show logged in state
Adds google icon to login button
Added license on logged-in page, but will add license to all files in a separate pr to minimize file changes


anncatton:
@ciaranschutte I didn't have time to look at getting rid of the `logged-in` page for fetching the ego token, but I do want to look at that option when I have a chance!

components/DefaultPageLayout.tsx
```diff
@@ -15,14 +15,21 @@ const DefaultPageLayout = ({
...
 }): ReactElement => {
   return (
     <ThreeRowPageContainer>
       <PageHead title={title} />
       <NavBar />
-      <main>{children}</main>
+      <main
+        css={css`
+          display: flex;
+          flex: 1;
```
caravinci, line 28: Are we able to apply this display mode to the individual page(s) that may need it, instead of globally?

Considering most markup relies on the browsers' default `block` or `inline` display layout (`flex` as is, forces non-wrapping row layouts), I believe this may break the flow in some cases/pages.

anncatton: I was wondering about that. I hadn't gotten this alignment working so far in the logged-in page - but i'll fiddle a bit more!

ciaranschutte (approved):

PR #90, opened 2021-06-02
✨ add 'manage applications' page

samrichca:
- add admin dashboard (just the UI in this ticket)
- add lodash & date-fns
- check for admin privileges on /applications route


global/constants/internalPaths.ts
```diff
@@ -1 +1 @@
-export const APPLICATIONS_PATH = 'applications';
+export const APPLICATIONS_PATH = '/applications';
```
samrichca, line 1: need the slash for links, doesn't break the existing usage in router.push() functions

caravinci: While I could be wrong, I think the reason you're needing those is because we may have no app root defined.
these constants should be `urlJoin` at the link upon said root. Maybe you're missing it in your `.env`?

samrichca: we don't have that parameter in this repo as far as i know, and we have a leading slash in internal paths in platform-ui so i followed that pattern

caravinci: oh, not a request for change at all... and didn't realise they were doing that there.
Hopefully I'll get a chance to get involved in that project, to get acquainted with how they do things there.

Jenkinsfile.groovy
```diff
@@ -77,7 +77,7 @@ spec:
         stage('install') {
           steps {
             container('node') {
-              sh "npm ci"
+              sh "npm ci --legacy-peer-deps"
```
caravinci, line 80: temporary change, while we figure out why it wasn't necessary earlier.

components/pages/Applications/ManageApplications/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type Accessors =
+  | 'appId'
+  | 'institution'
+  | 'applicant'
+  | 'googleEmail'
+  | 'ethicsLetter'
+  | 'accessExpiry'
+  | 'lastUpdated'
+  | 'status';
```
caravinci, line 20: think this can work as `export type Accessors = keyof Application;`

global/styles.ts
```diff
@@ -0,0 +1,9 @@
+import { css } from '@icgc-argo/uikit';
+
+export const instructionBoxButtonIconStyle = css`
+  margin-right: 5px;
+`;
+export const instructionBoxButtonContentStyle = css`
```
caravinci, line 6:
```suggestion
export const instructionBoxButtonContentStyle = css`

```

caravinci (approved):

PR #98, opened 2021-06-04
fix emotion uikit version dependency issues

caravinci:
Final summary of changes:

- Adds new UiKit version 1.10.4
- Adds the babel plugin for Emotion@10 (without removing the existing v11 one, which turned out useful) 
- Updates the nextjs module transpiler

(as a result of undoing the more dramatic initial commit)


babel.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
```
caravinci, line 1: Made this a `babel.config.js` in order to make it a project wide config, overriding other babel configs that may be embedded in any folder.

See for more info https://babeljs.io/docs/en/config-files

components/Footer.tsx
```diff
@@ -20,14 +19,14 @@ const { NEXT_PUBLIC_ARGO_ROOT, NEXT_PUBLIC_ARGO_PLATFORM_ROOT } = getConfig();
 
 const StyledLink = styled(Link)`
   ${({ theme }: { theme: UikitTheme }) => css`
-    ${theme.typography.data};
+    ${css`theme.typography.data`};
```
caravinci, line 22: this prevents a build issue that comes up from mixing the two @emotion libraries.
It wasn't coming up earlier because the code was Emotion@11 compliant.

ciaranschutte: ok fix for now! work ok.

ciaranschutte (approved):

caravinci:
@ciaranschutte I was able to REALLY simplify the fix, by keeping both the existing Emotion@11 babel plugin that Ann added (to build without nested `css`), while adding the v10 plugin to handle the labels created locally using uikit's `css`. Managed this by studying a couple of things we can improve at uikit itself later one when time allows.

---
TL;DR: this is not necessary anymore, if you use the `css` imported from uikit.  (https://github.com/icgc-argo/dac-ui/pull/98#discussion_r645840596)
image

PR #101, opened 2021-06-08
✨ add axios hook to connect to DAC-API

samrichca:
- add a proxy so we can make API requests from localhost UI
- add a hook for using axios to make ajax requests to DAC-API

this PR doesn't complete the linked ticket, it only provides basic axios/api functionality.


global/hooks/useAxios.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }).catch(err => {
+        setError(err);
+      }).finally(() => {
+        setLoading(false);
+      });
+  }
+
+  useEffect(() => {
+    fetchAPI();
+  }, []);
```
ciaranschutte, line 52: I think we need some dependencies of `useEffect` here. 
What if we pass in any new props to `useAxios`? right now it will just return the state without ever updating the state because state is only set `onMount`

ciaranschutte (changes requested):
looks great! One quick note about the `useEffect` dependencies in the hook

global/hooks/useAxios.tsx
caravinci: Agreed, @ciaranschutte. We can optionally get those from whichever component calls the hook.
Thinking it could be a value or an array of them passed, as a single props into the hook (e.g. `requestFor` or some better name), which are then put into that line 52.

samrichca: i have a design question about this actually, i'll ask you guys on slack in a bit

PR #104, opened 2021-06-09
✨ feat/manage applications data hookup

samrichca:
- add generic Axios function to interact with DAC-API
- add API proxy to deal with CORS issues
- add <ContentError /> component for API errors
- hookup "manage applications" page to API


global/hooks/useFetchManageApplications.tsx
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from 'react';
```
ciaranschutte, line 1: for naming, this is closer to a `useAxios` or a `useApi` as it'll be in use in other sections, not just 'ManageApplications'
logic looks great

components/pages/Applications/ManageApplications/utils.tsx
```diff
@@ -0,0 +1,106 @@
...
+    lastUpdated: datum.updatedAtUtc,
+    status: datum.state,
+  }));
+
+export const tableColumns: TableColumnConfig<ApplicationRecord> & {
+  id: ManageApplicationsField;
+} = [
+    {
+      Header: fieldDisplayNames['appId'],
+      id: ManageApplicationsField['appId'],
```
ciaranschutte, line 46: accessing an enum property that isn't a string literal can be done like: `ManageApplicationsField.appId` too :)

global/hooks/useFetchManageApplications.tsx
caravinci: how about `fetchApplicationData`?

ciaranschutte:
sorry I should have been more clear.
I just meant this one https://github.com/icgc-argo/dac-ui/pull/104/commits/f9daf021e097bfba4b6857f3276a639fc545da4c#diff-d1133ed1ffa36ac8e8861d98f521eb3f5f2d35e62fd0126c4728ffcb88611c33 seen as it's in the `global` folder.

this works too.

ciaranschutte (approved):

PR #108, opened 2021-06-14
Bug/hook in plain old func

ciaranschutte:
move authenticated fetch logic to `authContext` 
- fixes the case of not being able to use a hook in an ordinary js func
- this way we don't have to worry about grabbing the token for every request hook we write.


global/utils/index.ts
```diff
@@ -1 +1 @@
-export { default as fetchAPI } from './fetchAPI';
```
ciaranschutte, line 1: keeping file as it's good practice to use exports like this.

samrichca (approved):
nice improvements!

global/hooks/useAuthContext.tsx
```diff
@@ -9,20 +9,23 @@ import {
...
+  fetchWithAuth: any;
   permissions: string[];
 };
 
 const AuthContext = createContext<T_AuthContext>({
   token: undefined,
   logout: () => {},
   user: undefined,
-  fetchWithAuth: fetch,
+  fetchWithAuth: null,
```
anncatton, line 28: just set as `() => {}` so at least there's a function as default

global/hooks/useAuthContext.tsx
```diff
@@ -61,12 +64,36 @@ export const AuthProvider = ({
...
+
+  axios.defaults.baseURL = USE_DAC_API_PROXY ? '' : DAC_API;
+  axios.defaults.headers.post['Content-Type'] = 'application/json;charset=utf-8';
+  axios.defaults.headers.post['Access-Control-Allow-Origin'] = '*';
+
+  const fetchWithAuth = ({
+    params = {},
+    headers = {},
+    method = 'GET' as Method,
+    url = '/',
```
anncatton, line 77: might be better not to default this? i don't know if the request should be made if a valid endpoint is not provided

global/hooks/useAuthContext.tsx
```diff
@@ -61,12 +64,36 @@ export const AuthProvider = ({
...
-    return fetch(url, {
-      ...options,
-      headers: { ...options?.headers, accept: '*/*', Authorization: `Bearer ${token || ''}` },
-      body: null,
-    });
+  const { USE_DAC_API_PROXY } = getConfig();
+
+  axios.defaults.baseURL = USE_DAC_API_PROXY ? '' : DAC_API;
+  axios.defaults.headers.post['Content-Type'] = 'application/json;charset=utf-8';
+  axios.defaults.headers.post['Access-Control-Allow-Origin'] = '*';
```
anncatton, line 71: i forget how axios works...are these headers just slightly less generic than the headers the fetch function can take?

samrichca: should add a comment here in the code. i added these headers to enable a proxy so we don't have CORS errors when accessing DAC-API in local UI.

global/hooks/useAuthContext.tsx
samrichca: +1 for removing default URL

global/hooks/useAuthContext.tsx
anncatton: can we add them with the other defaults set in the `fetchWithAuth` function? just so it's clearer what the defaults are and where they're coming from. i.e. `headers: { post: {...defaultValues}, ...otherHeaders}`. Or does axios not like that...?

global/hooks/useAuthContext.tsx
ciaranschutte: :+1:

global/hooks/useAuthContext.tsx
samrichca: I'm not sure offhand but it's worth trying for clearer code

if the response comes back normally/without a CORS error then it worked

ciaranschutte: I'm going to have to touch base with you on the CORs stuff tomorrow (Tues) @samrichca

anncatton (approved):
thanks @ciaranschutte @samrichca !

PR #137, opened 2021-06-17
🔖 add version numbers to footer

samrichca:
- get app version from package.json
- get API from /health endpoint
- display both in footer


ciaranschutte (approved):
Looks good to me.

One change I'd suggest if it's straightforward with our current setup is to make the footer render once. Right now when you change pages, it appears Footer is re-rendering and is causing the `health` endpoint to be called again which leaves the text blank for a second.

samrichca:
> Looks good to me.
> 
> One change I'd suggest if it's straightforward with our current setup is to make the footer render once. Right now when you change pages, it appears Footer is re-rendering and is causing the `health` endpoint to be called again which leaves the text blank for a second.

duly noted, good catch!

Issue #51, opened 2021-05-26
[Data Hookup] Load Application and correct metadata (header)


kcullion:
## Goal
An application loads from the dashboard links or admin dashboard table in the correct state with the correct header metadata.

- included implementing the standard Data Provider (also in progress with @samrichca to be discussed)

## Endpoint to use
- powered by this endpoint: https://dac.qa.argo.cancercollaboratory.org/api/api-docs/#/Application/get_applications__id_ 

## Expected Outcomes
- [x] The correct application loads (landing on the introduction/application terms section): 
1. Clicking an application dashboard card
2. Clicking an application from the admin table dashboard
3. a direct URL is visited and the user has permission to load it (eg. from an email)

- [x] The header information has the expected information (some depends on filling out section A)
1. Application ID
2. Show the Creation date and last updated date/time
3. When applicant fills out the application name - show title, first name, last name in header bar
3. When applicant fills out primary affiliation in section A, show that with the applicant info too

![image](https://user-images.githubusercontent.com/18424138/120361779-43bb3980-c2d8-11eb-94b2-a1cba03a5d20.png)

- [x] Link "My Applications" back to the user's dashboard 
![image](https://user-images.githubusercontent.com/18424138/122234589-fec40500-ce8a-11eb-873b-a92c4d04aed0.png)

Progress bar data hookup is covered here: https://github.com/icgc-argo/dac-ui/issues/50

ciaranschutte:
as discussed in standup reassigning to @caravinci

caravinci:
Extending scope a smidge to include handling of this:
> https://github.com/icgc-argo/dac-api/issues/34#issuecomment-863479445
> ![image.png](https://images.zenhubusercontent.com/5defce3f99bcf0b1b8a98d24/cbf824b1-6ec3-41db-af38-0201121be83d)

rosibaj:
lgtm as a start.  thank you @caravinci!

PR #144, opened 2021-06-18
51 - improve the initial loads when logging in

caravinci:
Initial commit to reduce the size of my upcoming PR for #51, attempting to reduce/split complexity for the reviewing Dev Ninjas ⚔️

### DIFF
--- a/components/pages/Applications/index.tsx
+++ b/components/pages/Applications/index.tsx
@@ -1,5 +1,6 @@
 import { ReactElement } from 'react';
 import { useRouter } from 'next/router';
+import DnaLoader from '@icgc-argo/uikit/DnaLoader';
 
 import { isDacoAdmin } from 'global/utils/egoTokenUtils';
 import { useAuthContext } from 'global/hooks';
@@ -19,7 +20,7 @@ const Application = (): ReactElement => {
   const {
     query: { ID: [appId = ''] = [] },
   }: QueryType = useRouter();
-  const { permissions } = useAuthContext();
+  const { loadingAuth, permissions } = useAuthContext();
 
   const isAdmin = isDacoAdmin(permissions);
 
@@ -27,7 +28,9 @@ const Application = (): ReactElement => {
 
   return (
     <DefaultPageLayout title={pageTitle}>
-      {appId ? (
+      {loadingAuth ? (
+        <DnaLoader />
+      ) : appId ? (
         <ApplicationForm appId={appId} isAdmin={isAdmin} />
       ) : isAdmin ? (
         <ManageApplications />
--- a/global/constants/externalPaths.ts
+++ b/global/constants/externalPaths.ts
@@ -14,7 +14,6 @@ export const CONTACT_PAGE = urlJoin(NEXT_PUBLIC_ARGO_PLATFORM_ROOT, 'contact');
 export const POLICIES_PAGE = NEXT_PUBLIC_ARGO_ROOT; // tbd
 export const HELP_PAGE = NEXT_PUBLIC_ARGO_DOCS_ROOT; // tbd
 export const CONTROLLED_DATA_USERS_PAGE = NEXT_PUBLIC_ARGO_ROOT; // tbd
-export const DAC_API = NEXT_PUBLIC_DAC_API_ROOT;
 
 export const PRIVACY_POLICY_PAGE = urlJoin(ARGO_POLICIES, '2/privacy');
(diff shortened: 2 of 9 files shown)


components/pages/Applications/index.tsx
```diff
@@ -19,15 +20,17 @@ const Application = (): ReactElement => {
...
 
   const isAdmin = isDacoAdmin(permissions);
 
   const pageTitle = appId.toUpperCase() || 'Application page';
 
   return (
     <DefaultPageLayout title={pageTitle}>
-      {appId ? (
+      {loadingAuth ? (
+        <DnaLoader />
```
caravinci, line 32: add a loader for the `Applications` base layout, while the auth sorts itself out.

global/constants/externalPaths.ts
```diff
@@ -14,7 +14,6 @@ export const CONTACT_PAGE = urlJoin(NEXT_PUBLIC_ARGO_PLATFORM_ROOT, 'contact');
 export const POLICIES_PAGE = NEXT_PUBLIC_ARGO_ROOT; // tbd
 export const HELP_PAGE = NEXT_PUBLIC_ARGO_DOCS_ROOT; // tbd
 export const CONTROLLED_DATA_USERS_PAGE = NEXT_PUBLIC_ARGO_ROOT; // tbd
-export const DAC_API = NEXT_PUBLIC_DAC_API_ROOT;
```
caravinci, line 17: unused variable

global/hooks/useApplicationsAPI.tsx
```diff
@@ -22,29 +22,30 @@ ApplicationsRequestData) => {
...
+        .catch((err: any) => {
+          setError(err);
+        })
+        .finally(() => {
+          setIsLoading(false);
+        });
   }, [page, pageSize, sort]);
 
-  return { error, isLoading, response };
+  return { error, isLoading: loadingAuth || isLoading, response };
```
caravinci, line 48: get the loading state from the auth context provider, and prioritise it over the fetcher's, as no calls should be made while there's no valid token.

global/hooks/useApplicationsAPI.tsx
```diff
@@ -22,29 +22,30 @@ ApplicationsRequestData) => {
...
-      params: {
-        page,
-        pageSize,
-        sort,
-      },
-      url: API.APPLICATIONS,
-    })
-      .then((res: any) => {
-        setResponse(res);
+    token &&
```
caravinci, line 28: ensure no auth call is made if there's no token at all.
some calls may use an invalid token, but that's a different scenario.

global/hooks/useAuthContext.tsx
```diff
@@ -50,18 +54,16 @@ export const AuthProvider = ({
...
+  if (token) {
     if (!isValidJwt(token)) {
       if (egoJwt && token === egoJwt) {
         removeToken();
       }
     } else if (!egoJwt) {
       setTokenState(null);
     }
+  } else if (isValidJwt(egoJwt)) {
+    setTokenState(egoJwt);
```
caravinci, line 66: do not not ask positives whenever possible, because life is easier this way.

global/hooks/useAuthContext.tsx
```diff
@@ -91,20 +93,20 @@ export const AuthProvider = ({
...
+      // TODO log errors somewhere?
+      console.error({ error });
+    });
   };
 
   const userInfo = token ? decodeToken(token) : null;
   const user = userInfo ? extractUser(userInfo) : undefined;
   const permissions = getPermissionsFromToken(token);
 
+  loadingAuth && token && user && setLoadingAuth(false);
```
caravinci, line 106: this ensures the context provider remains in loading state if there's unfinished/unhandled business.

ciaranschutte:
I'm wondering are we missing anything in the Next way of doing things? Should we maybe be using some of their life cycle methods like `getInitialProps`?

caravinci:
Great idea! I'll look into that option, to see what can be done with them that may help us; while this should improve the login experience while we find a more `NextJS`y solution (if there's any)

caravinci:
Updated this upon Sam's recent changes @ Management.
After some due research and team conversation, we decided it'd be best to not implement these changes using SSR for the time being, as that would require some deeper (aka out of scope) changes.

This PR should be ok to review now.

ciaranschutte (approved):
Looks good to me. Inline with all the offline discussions.

PR #150, opened 2021-06-23
footer API version: use axios instead of fetchWithAuth

samrichca:
using axios directly for the health API endpoint because it's our only no-auth API request

### DIFF
--- a/global/hooks/useHealthAPI.tsx
+++ b/global/hooks/useHealthAPI.tsx
@@ -1,23 +1,21 @@
 import { useEffect, useState } from 'react';
-import { AxiosError, AxiosResponse } from 'axios';
-import useAuthContext from './useAuthContext';
+import axios, { AxiosError, AxiosResponse } from 'axios';
 import { API } from 'global/constants/externalPaths';
+import { getConfig } from 'global/config';
+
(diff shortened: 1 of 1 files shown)


global/hooks/useHealthAPI.tsx
```diff
@@ -1,23 +1,21 @@
...
-  const { fetchWithAuth } = useAuthContext();
-
   useEffect(() => {
-    fetchWithAuth({
-      url: API.HEALTH,
-    })
-      .then((res: any) => {
+    axios.get(API.HEALTH, { baseURL: NEXT_PUBLIC_DAC_API_ROOT })
+      .then((res: AxiosResponse | undefined) => {
         setResponse(res);
```
anncatton, line 16: can we check the response is ok before setting the response (and throw error if it is not)?

samrichca: good idea, will do

samrichca: added `validateStatus` to axios

anncatton: oooh fancy

anncatton (approved):

Issue #128, opened 2021-06-17
PDF: Render dynamic user data into pdf


rosibaj:
- get the app data for this directly from the api (so that part of the pdf won’t rely on what the ui is doing). 
- render the state of the users selections on the pdf template.

anncatton:
@rosibaj @kcullion Known issues from this PR:
- `IT Agreements` and `Data Access Agreements` sections are missing some of the grey borders in the web ui
- the loading state for the PDF download should not resize the button (this happens because the button text changes)
- the PDF download button text is hard coded to `Draft` and does not reflect the current app state

kcullion:
thanks @anncatton. do you suggest we make tickets for all of these or will you be fixing as part of this ticket?

anncatton:
I just added this as a comment on the ticket so you and Rosi are aware. I don't think any new tickets should be made until you've had a chance to QA everything on this ticket, and then we can go from there.

kcullion:
@anncatton @rosibaj I put Ann's second and third point in this new ticket: https://github.com/icgc-argo/dac-ui/issues/181
I put the fourth item in this ticket so we remember to test it: https://github.com/icgc-argo/dac-ui/issues/175 

Rosi please review that third item about the signed pdf download - what was decided - are we zipping or including ethics letter in pdf?

kcullion:
@anncatton Some more feedback 

- [ ] fix the url, change the wording and remove the dash from ICGC-DACO `DACO-70 created for <applicant title><applicant first name><applicant last name> by ICGC DACO (https://daco.icgc-argo.org)` and if there is no applicant name filled in yet, then include  <applicant name> in the string

rosibaj:
Tested Jul 07, 2021: 
- [ ] some ui issues with the publication display
![image](https://user-images.githubusercontent.com/15824252/124793966-83e78a80-df1c-11eb-8d8a-54b2487e1a05.png)

- [ ] lay summary is not being included and should be 
![image](https://user-images.githubusercontent.com/15824252/124794095-a7aad080-df1c-11eb-8ba1-ae4c412d52de.png)
![image](https://user-images.githubusercontent.com/15824252/124794038-98c41e00-df1c-11eb-99d0-96a9b1748280.png)


@kcullion the zip package includes the ethics letter when it is signed. that should be delivered fine by the api rn

kcullion:
[...]

Issue #127, opened 2021-06-17
PDF: Static text between app + pdf


rosibaj:
- dont want to duplicate text between the ui pages and pdf creating a copy-paste nightmare
- I’m working on a setup to have all the static text shared between the ui and the pdf

anncatton:
Initial setup in `0.9.0`:
- hookup pdf download to `Draft PDF` button (button does not reflect app state yet)
- pdf will show static text for `Introduction` and `Applicant Information`
- `Introduction` acknowledgement should show correct state from what is latest in the api
- watermark should reflect current app state
- `Applicant Information` has only static text and form data subtitle, no app state yet

Work for the rest of the pdf is continuing and will appear in a later release

Issue #60, opened 2021-05-26
[Data-Hookup] Generate pdf with applicant data in correct states


kcullion:
- always export CURRENT state of the form, in the pdf
- Has draft watermark if in draft state, otherwise no watermark
- test with 1 section downloading the values based on the investigated method. 
- (Not sure where to put this) - for all links in the content of the form, display the url in brackets after that link text within the pdf

UI zeplin links can be found: https://github.com/icgc-argo/dac-ui/issues/64

anncatton:
Initial setup in 0.9.0:

- hookup pdf download to Draft PDF button (button does not reflect app state yet)
pdf will show static text for `Introduction` and `Applicant Information`
- `Introduction` acknowledgement should show correct state from what is latest in the api
- watermark should reflect current app state
- `Applicant Information` has only static text and form data subtitle, no app state yet

Work for the rest of the pdf is continuing and will appear in a later release

kcullion:
thanks @anncatton . Will review more when more sections are in there but for now, let's change the name format to 
`ICGC-DACO-32-20210628.pdf`

PR #153, opened 2021-06-24
Static components for PDF

anncatton:
WIP for PDF feature. On recommendation from @blabadi , getting this portion into dev sooner rather than later to kickstart feedback
- adds components for static text portions of `Introduction` and `Applicant` sections
- adds some simple styled components for pdf display
- generates pdf on demand. api call for app data made in this onClick function so data is most recently saved state
- setup for form text still *very* wip, as i haven't completely decided on the best approach for this (so feedback is very welcome!). PDF text for the form inputs is not always the same as the UI
- have added only a few data bits to the pdf to test it is working (applicant info, introduction acknowledgement)
- will be adding type info as work progresses
- addresses portions of #127 , #128 and #60


components/Head.tsx
```diff
@@ -5,7 +5,7 @@ const Head = () => {
   return (
     <NextHead>
       <link
-        href={'https://fonts.googleapis.com/css?family=Lato:300,400,600&display=swap'}
+        href={'https://fonts.googleapis.com/css?family=Work+Sans:300,400,600&display=swap'}
```
samrichca, line 8: just checked, daco has this too. i'll correct it on my next PR :P

anncatton: yeah i forgot to note that in the pr description...i believe this was my mistake! head still in overture space 🙂

components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -21,8 +42,21 @@ const HeaderActions = (): ReactElement => {
...
-
-      <Button onClick={function noRefCheck() {}} size="sm" variant="secondary">
+      <Button
+        isAsync
+        variant="secondary"
+        size="sm"
+        onClick={async () => {
+          const data = await fetchWithAuth({ url: urlJoin(APPLICATIONS_PATH, appId) })
+            .then((res: any) => {
+              if (res.status === 200) {
```
samrichca, line 52: I'm gonna add validateStatus 200 to fetchWithAuth, so the status check can be removed here

components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -21,8 +42,21 @@ const HeaderActions = (): ReactElement => {
...
+                return res.data;
+              }
+            })
+            .catch((err: any) => console.log('ERROR! ', err));
```
samrichca, line 56: console.warn instead?

anncatton: yes ma'am! forgot about this one, nice catch :)

samrichca (approved):
this pdf solution looks really good!! nice work

components/pages/Applications/ApplicationForm/Header/Actions.tsx
samrichca: catching it again........ we're using console.error in other fetch errors... my b

blabadi (approved):

caravinci (approved):
Looks great to me 🚀

PR #154, opened 2021-06-24
add appId to useApplicationsApi

caravinci:
using the Application header as a test subject, @samrichca pointed out the hook was missing the capacity to pass data body and methods other than GET. This PR addresses that, as well as covering some technical debt I had introduced in the header (which account for the extra changes).

### DIFF
--- a/components/Loader.tsx
+++ b/components/Loader.tsx
@@ -1,0 +1,18 @@
+import { ReactElement } from 'react';
+import { css } from '@emotion/core';
+import DnaLoader from '@icgc-argo/uikit/DnaLoader';
+
+const Loader = (): ReactElement => (
+  <figure
+    css={css`
+      align-items: center;
+      display: flex;
+      height: 100%;
+      justify-content: center;
+    `}
+  >
+    <DnaLoader />
+  </figure>
+);
+
+export default Loader;
--- a/components/pages/Applications/ApplicationForm/index.tsx
+++ b/components/pages/Applications/ApplicationForm/index.tsx
@@ -1,31 +1,22 @@
 import { ReactElement, useEffect, useState } from 'react';
 
-import { API } from 'global/constants';
-import { useAuthContext } from 'global/hooks';
+import Loader from 'components/Loader';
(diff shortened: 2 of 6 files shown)


components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -11,18 +11,11 @@ import DnaLoader from '@icgc-argo/uikit/DnaLoader';
...
-      url: `${API.APPLICATIONS}/${appId}`,
-    })
-      .then(({ data }: { data: Record<string, any> }) => setAppData(data))
-      .catch((error: Error) => {
-        // TODO dev logging, errors should not be shown to user
-        console.error(error);
-      });
-  }, []);
+    response && setAppData(response.data);
+  }, [response]);
```
caravinci, line 18: Preventing technical debt from using the wrong hook

global/hooks/useApplicationsAPI.tsx
```diff
@@ -34,7 +35,7 @@ ApplicationsRequestData) => {
           sort: stringifySort(sort),
           states: stringifyStates(states),
         },
-        url: API.APPLICATIONS,
+        url: `${API.APPLICATIONS}/${appId}`,
```
caravinci, line 38: this handles both having and not the appId available

components/Loader.tsx
```diff
@@ -0,0 +1,18 @@
+import { ReactElement } from 'react';
+import { css } from '@emotion/core';
+import DnaLoader from '@icgc-argo/uikit/DnaLoader';
+
+const Loader = (): ReactElement => (
```
caravinci, line 5: new self centring, very reusable component.

components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -1,31 +1,22 @@
...
 
   return isLoading || Object.values(appData).length < 1 ? (
-    <DnaLoader />
+    <Loader />
```
caravinci, line 19: better loader, because <3

samrichca (approved):

PR #156, opened 2021-06-25
fixes for fetching API

samrichca:
- default page size is now 20 on Manage Applications page
- renamed useApplicationsAPI to useGetApplications
- rename statesAllowList to adminStatesAllowList, to ensure we can share constants/types between components
- used URL join instead of string concatenation when creating a new application
- changed logout:
  - add session_expiry=true to URL
  - logout when fetchWithAuth happens after your token expires


global/hooks/useAuthContext.tsx
```diff
@@ -52,13 +52,13 @@ export const AuthProvider = ({
...
     removeToken();
-    router.push('/');
+    router.push('/?session_expired=true');
   };
 
   if (token) {
     if (!isValidJwt(token)) {
       if (egoJwt && token === egoJwt) {
-        removeToken();
+        logout();
```
samrichca, line 61: testing fixing @anncatton 's PDF API call woes

global/hooks/useGetApplications.tsx
```diff
@@ -11,43 +12,48 @@ import {
...
-        });
+        })
+        .finally(() => {
+          setIsLoading(false);
+        })
     }
-  }, [page, pageSize, stringifySort(sort)]);
+  }, [appId, page, pageSize, stringifySort(sort)]);
 
   return { error, isLoading, response };
```
samrichca, line 56: this was passing down token loading status not API loading status

ciaranschutte (approved):

PR #158, opened 2021-06-28
Pdf dynamic data

anncatton:
PDF setup with static and dynamic content. Addresses #127 #128 and #60 
- adds components for all static text sections, where possible
- adds text constants for portions of the Application form sections, used in UI and PDF components
- api request will be made on demand, when the download pdf button is clicked, so app data is up-to-date and independent of the ui state
- typing and enums for Application data
- there are some minor styling issues related to the css on `section` elements in the ui, but I think it's best to get this feature into testing as soon as can be (and if someone sees the solution in review, that also works!)


components/pages/Applications/ApplicationForm/Forms/EthicsLetter.tsx
```diff
@@ -86,20 +54,16 @@ const EthicsLetter = ({
...
             <FormRadio disabled={isSectionDisabled} value="false" checked>
-              You represent and warrant that your country/region does not require your Research
-              Project to undergo ethics review.
+              {FORM_TEXT.ethics.declarationOptions.notRequired}
             </FormRadio>
             <FormRadio disabled value="true">
-              Your country/region requires your Research Project to undergo ethics review, and
-              therefore, this Research Project has been approved by an IRB/REC formally designated
-              to approve and/or monitor research involving humans.{' '}
+              {FORM_TEXT.ethics.declarationOptions.required.a}{' '}
```
anncatton, line 60: not a huge fan of the naming convention i used here, i'm open to suggestions!

components/pages/Applications/ApplicationForm/Forms/Introduction.tsx
```diff
@@ -25,11 +37,20 @@ const Introduction = ({
   validateFieldTouched: FormFieldValidationTriggerFunction;
 }): ReactElement => {
   // for local testing, will be removed when pdf feature complete
-  const [isClient, setIsClient] = useState(false);
-  useEffect(() => {
-    // cannot render PDFDownloadLink on server side, dynamically importing did not resolve the issue
-    setIsClient(true);
-  }, []);
+  const { fetchWithAuth } = useAuthContext();
```
anncatton, line 40: will remove this request and all the pdf components once pr review is finished; is for local testing only

components/pages/Applications/ApplicationForm/Forms/types.ts
```diff
@@ -97,7 +98,7 @@ export type FormSectionValidationState_Applicant = FormSectionValidationState_Se
   address_cityAndProvince: { value: string };
   address_country: { value: string };
   address_street: { value: string };
-  address_postalCod: { value: string };
+  address_postalCode: { value: string };
```
anncatton, line 101: assuming this is a typo :)

components/pages/Applications/ApplicationForm/Forms/types.ts
```diff
@@ -146,24 +147,22 @@ export type FormSectionValidationState_ProjectInfo =
...
-    info_middleName: { value: string };
-    info_positionTitle: { value: string };
-    info_primaryAffiliation: { value: string };
-    info_suffix: { value: string };
-    info_title: { value: string };
-    address_building: { value: string };
-    address_cityAndProvince: { value: string };
-    address_country: { value: string };
-    address_street: { value: string };
-    address_postalCod: { value: string };
```
anncatton, line 163: just typo fix

components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -24,10 +33,18 @@ const HeaderActions = ({ appId }: { appId: string }): ReactElement => {
       <Document>
```
anncatton, line 33: PDF will be created on demand, instead on on mount

components/pages/Applications/PDF/PdfLayout.tsx
```diff
@@ -10,7 +10,7 @@ const styles = StyleSheet.create({
     position: 'relative',
     flexDirection: 'column',
```
anncatton, line 11: basic pdf layout, likely styles will change after QA

components/pages/Applications/PDF/StaticAppendices.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { AppendixAgreement, ApplicationData } from '../types';
+
+export const ICGCPolicies = ({ isPdf = false }: { isPdf?: boolean }) => {
+  const { SectionTitle } = getStaticComponents(isPdf);
+  return <SectionTitle>ICGC POLICIES</SectionTitle>;
+};
+
+const Appendix = ({ agreement }: { agreement: AppendixAgreement }) => {
+  return (
+    <View style={{ marginBottom: '10pt' }} wrap={false}>
```
anncatton, line 17: `wrap=false` prevents the component from flowing into the next page

components/pages/Applications/PDF/StaticApplicant.tsx
```diff
@@ -1,11 +1,63 @@
...
-import { getStaticComponents, SectionTitle } from './common';
+import { getStaticComponents, PdfFormFields, SectionTitle } from './common';
 import FORM_TEXT from './textConstants';
+import VerticalTable from './VerticalTable';
+import { getStreetAddress } from './common';
+import { ApplicationData } from '../types';
 
-const StaticApplicant = ({ isPdf = false, data = {} }: { isPdf?: boolean; data?: any }) => {
+const PdfApplicantFormData = ({ data }: { data: any }) => {
+  const applicantFields = [
```
anncatton, line 13: The PDF tables all have *slightly* different fields to be displayed, so building each one with the available field enums

components/pages/Applications/PDF/StaticDataAccessAgreement.tsx
```diff
@@ -0,0 +1,393 @@
...
+    ContainerComponent,
+    SectionComponent,
+    TextComponent,
+    TitleComponent,
+    SectionTitle,
+    LinkComponent,
+    ListComponent,
+  } = getStaticComponents(isPdf);
+
+  const ListAsTypography = isPdf ? View : Typography;
```
anncatton, line 67: Cannot use `TextComponent` in this case as react-pdf `Text` component doesn't work well as a list container, but I wanted to keep `Typography` on the ui side as this works much better for styling.

components/pages/Applications/PDF/StaticDataAccessAgreement.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+  const termsAndConditions = [
+    ({ count }: { count: number }) => (
```
anncatton, line 70: Makeshift numbering for PDF ordered list.

components/pages/Applications/PDF/StaticDataAccessAgreement.tsx
```diff
@@ -0,0 +1,393 @@
...
+      >
+        <SectionTitle>DEFINITIONS</SectionTitle>
+
+        <ListAsTypography
+          component="ul"
+          css={css`
+            border: 1px solid ${defaultTheme.colors.grey_2};
+            list-style: none;
+            padding: 15px;
+            li {
```
anncatton, line 221: moved some css here as it was not applying from outside the component, looks like I have changed the inheritance a little but I didn't want to start changing styles external to this.

components/pages/Applications/PDF/StaticITAgreements.tsx
```diff
@@ -0,0 +1,229 @@
...
+        you and the cloud provider. To the extent permitted by law OICR accepts no responsibility
+        for any interaction between you and the cloud provider and is released from any liability
+        arising out of or in any way connected with such interaction.
+      </ListComponent>
+    ),
+  ];
+
+  // cannot use TextComponent in pdf context, cannot use plain <ul> in ui context
+  // is there a nicer way to do this
+  const UlAsTypography = isPdf ? View : Typography;
```
anncatton, line 138: same case as earlier comment about list styling in pdf

components/pages/Applications/PDF/StaticProjectInfo.tsx
```diff
@@ -0,0 +1,343 @@
...
+    TextComponent,
+  } = getStaticComponents(isPdf);
+
+  return (
+    <GenericContainer style={styles.bubbleContainer} wrap={false}>
+      <TextComponent as="span" style={{ margin: 0 }}>
+        Provide a short summary of the background basis of your research. For example,
+      </TextComponent>
+      <UnorderedListComponent>
+        <ListComponent asListItem style={{ width: '95%' }}>
```
anncatton, line 56: the use of a bullet (or another kind of element) in lists seems to push the entire text element over, and sometimes will render outside the list container. Decreasing width to account for this.

components/pages/Applications/PDF/StaticProjectInfo.tsx
```diff
@@ -0,0 +1,343 @@
...
+          <LinkComponent href="#" rel="noopener noreferrer" target="_blank">
+            Term 5 of the Data Access Agreement (DAA)
+          </LinkComponent>
+          , you agree not to link or combine the ICGC Controlled Data to other data available in a
+          way that could re-identify the Research Participants.{' '}
+          <TextComponent as="span" bold style={{ fontWeight: 600 }}>
+            Please confirm how the methods you intend to use to combine datasets will minimize the
+            risk for re-identification of the ICGC data donors
+          </TextComponent>
+          <TextComponent as="span" style={{ margin: 0 }}>
```
anncatton, line 129: ALL text needs to be wrapped in a `Text` component in react-pdf, or rendering errors occur.

components/pages/Applications/PDF/StaticProjectInfo.tsx
```diff
@@ -0,0 +1,343 @@
...
+            {FORM_TEXT.project_info.inputLabel.dataUse}
+          </PDFParagraph>
+          <DataUseBubble isPdf />
+          <PDFTextArea>{data?.sections.projectInfo.methodology}</PDFTextArea>
+          <StaticLaySummary isPdf />
+          <PDFParagraph style={{ fontWeight: 600 }}>
+            {FORM_TEXT.project_info.inputLabel.laySummary}
+          </PDFParagraph>
+          <LaySummaryBubble isPdf />
+          {/* TODO: add lay summary from data */}
```
anncatton, line 334: Did not see `laySummary` in api data?

components/pages/Applications/PDF/StaticProjectInfo.tsx
```diff
@@ -0,0 +1,343 @@
...
+      </SectionComponent>
+      {isPdf && (
+        <View>
+          <BasicInfo data={data?.sections.projectInfo} />
+          <StaticResearchSummary isPdf={isPdf} />
+          <PDFParagraph style={{ fontWeight: 600 }} wrap>
+            {FORM_TEXT.project_info.inputLabel.background}
+          </PDFParagraph>
+          <BackgroundBubble isPdf />
+          <PDFTextArea>{data?.sections.projectInfo.background}</PDFTextArea>
```
anncatton, line 318: Because we have only a word limit, not a char limit, text areas in pdf are just View containers that will render only as much space as is needed for the tex.

components/pages/Applications/PDF/Table.tsx
```diff
@@ -0,0 +1,89 @@
...
+  },
+  tableCellStyle: {
+    padding: '6pt',
+    fontSize: 11,
+    fontFamily: 'WorkSans',
+  },
+});
+
+// needs to be generalized; right now only works for approval letters table
+const Table = ({ headers, data }: { headers: { name: string; accessor: string }[]; data: any }) => {
```
anncatton, line 29: This component was meant to be a reusable Horizontal table, but I am blanking on setting the data accessors up. So for now, it is used only for the ethics letter file upload table (all other tables use VerticalTable).

components/pages/Applications/PDF/textConstants.ts
```diff
@@ -4,6 +4,106 @@ const FORM_TEXT = {
...
+        link: `Data Access Agreement`,
+        c: `, current and applicable ethical approval is the responsibility of the Principal Investigator.`,
+      },
+    },
+  },
+  itAgreements: {
+    commaSeparator: ', ',
+    yes: 'Yes',
+    declarations: {
+      it_agreement_software_updates: `You will keep all computer systems on which ICGC Controlled Data reside, or which provide access to such data, up-to-date with respect to software patches and antivirus file definitions (if applicable).`,
```
anncatton, line 52: tried to use api keys where possible for consistency.

components/pages/Applications/types.ts
```diff
@@ -68,6 +73,143 @@ export type ApplicationsResponseItem = {
   state: string;
 };
 
+interface Address {
```
anncatton, line 76: Typing for ApplicationData + sections. It's possible I've missed some types that already exist, and some of this can probably be extended from others or used in more places.

components/pages/Applications/ApplicationForm/Forms/types.ts
ciaranschutte: i think so. seems fishy

ciaranschutte (approved):
Looks really great! Tonne of work here!
Builds, type-checks. I clicked the button and - blamo - beautiful pdf Thanks @anncatton !

components/pages/Applications/ApplicationForm/Forms/Introduction.tsx
```diff
@@ -1,19 +1,31 @@
...
 
 import StaticIntroduction from 'components/pages/Applications/PDF/StaticIntroduction';
-import FORM_TEXT from 'components/pages/Applications/PDF/textConstants';
 
 import {
   FormFieldValidationTriggerFunction,
   FormSectionValidationState_Introduction,
 } from './types';
 import { isRequired } from './validations';
+import { css } from '@icgc-argo/uikit';
```
caravinci, line 15: The @emotion versions mismatch created a few issues earlier in the project. UiKit's is outdated.
Interested in this because I'm relying on this for the layout of other components, so I want to resolve any issues my using the locally installed package may have caused in your implementation.

components/pages/Applications/ApplicationForm/Forms/types.ts
caravinci: Thanks for catching that!

caravinci (approved):
Looking good, looking good!

components/pages/Applications/ApplicationForm/Forms/Appendices.tsx
```diff
@@ -60,13 +52,13 @@ const Appendices = ({
...
           <FormCheckbox
-            aria-label="You have read APPENDIX I"
+            aria-label={FORM_TEXT.appendices.appendix_icgc_goals_policies.text}
             checked={localState.agreements?.fields?.appendix_icgc_goals_policies?.value}
             onBlur={validateFieldTouched}
             onChange={validateFieldTouched}
             value="agreements--appendix_icgc_goals_policies"
           >
-            You have read APPENDIX I
+            {FORM_TEXT.appendices.appendix_icgc_goals_policies.text}
```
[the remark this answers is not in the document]
anncatton: Yeah that sounds like a good direction to move in...i would feel more confident having these broken into components rather than the dictionary setup i have here!

components/pages/Applications/ApplicationForm/Forms/Introduction.tsx
anncatton: I just defaulted to the uikit version. I didn't have any issues during the pdf work, but i do remember the mismatch problems when i initialized this project...i can switch this back to the emotion import if it will cause problems!

caravinci: If you ask for my personal opinion, in general terms, importing a third party library from inside another dependency is an anti-pattern; of course unless said dependency was designed as a plugin/extension of the other.

To illustrate what I mean:
- bad: importing `jquery` from (some imaginary export given by) `bootstrap`.
- ok: importing `moment` from `moment-timezone`.

Apart from the long-winded build issues I stop-gapped, binding to that old version means we're not able to update and resolve vulnerabilities. Thus, the least we code now relying on it, the easier it'll be to keep up to date (particularly as the @emotion API and dependency tree have changed a little).

anncatton: Yeah, I don't disagree with that logic. My impression was the reason we had been importing `css` from uikit when uikit was being used, was because of that emotion version mismatch. But I could be wrong on the reasoning! Anyway, I will switch this back, but I think we probably need to decide as a team how we want to approach this, as I think we have a mix of both in ARGO.

anncatton: Oh actually in this particular file, nvmd...I just erased the import altogether cuz it was just in my testing code :)

caravinci: While I agree 100% on talking with the team about this, I'm not part of that app, so I have little to no say on how they should approach this. However if I recall correctly, Platform-UI imports the UiKit components directly as part of the same app, rather than as an imported package;  and so, both get built at the same time using one same @emotion version.

Leaving aside my opinion of the "anti-pattern" (subjective, I know), I didn't have time to test whether it makes any difference importing @emotion directly or from UiKit in that app's case; but I'd wager as they both build together upon deployment, it may not matter.

Definitely something worth exploring for future UiKit versions though, if we intend it to be reusable as a package.

caravinci: Oh and if I'm not mistaken, I think the binding to that internal version is because of the way Emotion handles the theme Provider's context (at least in older versions, which I believe they improved in the latest updates. Haven't gotten a chance to try it out.) 
Mixing versions gives issues because of the way the theme is passed to the dependent components, requiring additional theme providers for each version, etc. So it was easier to just share the single one.

Issue #168, opened 2021-06-29
🐛 Application fields - submitted in postman but there are UI issues 


kcullion:
## Describe the bug
The following fields need to be fixed

- [x] When I entered a Street Address in postman (section A) it didn't come through on the UI (

- [x] Section D: I see the project website value in postman, but it doesn't come through on the UI. 
- [x] FOR TESTING PURPOSES It needs to use the same validation as the other URLs because when I add a url that isn't the right format, it currently does not show an error icon in the Table of contents. 

![image](https://user-images.githubusercontent.com/18424138/123844238-c687e200-d8e0-11eb-8b97-36086ddf37b2.png)

![image](https://user-images.githubusercontent.com/18424138/123844320-def7fc80-d8e0-11eb-8bc8-f6e6b715c7c7.png)

- [ ] Institutional Website - this wasn't coming through on the UI

rosibaj:
must be merged/deployed with https://github.com/icgc-argo/dac-api/issues/71

PR #169, opened 2021-06-30
168 - field name corrections

caravinci:
Addresses feedback in #168

### DIFF
--- a/components/pages/Applications/index.tsx
+++ b/components/pages/Applications/index.tsx
@@ -24,14 +24,15 @@ const Application = (): ReactElement => {
   const { isLoading, permissions } = useAuthContext();
 
   const isAdmin = permissions.length > 0 && isDacoAdmin(permissions);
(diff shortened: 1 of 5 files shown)


components/pages/Applications/index.tsx
```diff
@@ -24,14 +24,15 @@ const Application = (): ReactElement => {
   const { isLoading, permissions } = useAuthContext();
 
   const isAdmin = permissions.length > 0 && isDacoAdmin(permissions);
-  const pageTitle = appId.toUpperCase() || 'Application page';
+  const normalisedAppId = appId.toUpperCase();
```
caravinci, line 27: Make sure the URL converts lowercased Application IDs, as the API is case sensitive.

caravinci (commented):
Commentssssss... 🐍

ciaranschutte (approved):

PR #182, opened 2021-07-07
161 - enable application autosave functionality

caravinci:
As described, these changes add persistence to (most of) the application fields.
Still missing: Collaborators, Ethics, and Sign+Submit (all of which rely on other endpoints, not PATCH.)


components/pages/Applications/ApplicationForm/Forms/Outline/index.tsx
```diff
@@ -46,16 +46,14 @@ const Outline = ({
...
         return (
           <FormSection
             active={selectedSection === name}
             key={name}
             label={sectionsData[name]?.description || name}
-            status={status}
+            status={formState.sections[name]?.meta.overall}
             switchSection={() => setSelectedSection(name)}
-            tooltip={sectionsData[name]?.tooltips?.[status] || ''}
+            tooltip={sectionsData[name].tooltips?.[status] || ''}
```
caravinci, line 56: all sections have implementation in the `sectionsData` object (even if they don't actually have `tooltips`, which is handled accordingly.)

caravinci (commented):
The usual mumbo-jumbo 🐘

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -256,24 +231,72 @@ export const validator: FormSectionValidatorFunction_Main =
...
+                ),
+              },
+              // __v: formState.__v,
+            },
+          }).then(({ data, ...response } = {} as AxiosResponse<any>) => {
+            data
+              ? dispatch({
+                  type: 'updating',
+                  value: data,
+                })
```
caravinci, line 258: Comment re-added because it got "outdated" by amending a bug I noticed:

> this function makes the call to persist the values, and then dispatches the new application state, to be reflected in the UI

components/pages/Applications/ApplicationForm/Forms/validations/helpers.ts
```diff
@@ -104,6 +123,71 @@ const getSeedValueByFieldType = (fieldType: string, fieldBase: any, seedValue: a
...
+  const [fieldName, fieldNameInner] = fieldNameSource.split('_');
+
+  switch (type) {
+    case 'array':
+      const meep = Object.entries({
+        ...fieldValue,
+        ...value,
+      });
+      return {
+        [fieldName]: meep
```
caravinci, line 141:
```suggestion
      return {
        [fieldName]: Object.entries({
          ...fieldValue,
          ...value,
        })
```

components/pages/Applications/ApplicationForm/Forms/types.ts
```diff
@@ -59,15 +61,19 @@ export type FormFieldType = Partial<
...
+        type: 'boolean';
+      };
     }
 >;
 
-type FormSectionValidationState_SectionsGenericType<
-  T extends Record<string, FormFieldType>
-> = Record<string, Partial<T[keyof T] & FormFieldType>>;
+type FormSectionValidationState_SectionsGenericType<T extends Record<string, FormFieldType>> =
+  Record<string, Partial<T[keyof T] & FormFieldType>>;
```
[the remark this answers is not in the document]
ciaranschutte: doesn't look like we have a prettier config in the project so itll use our individual default ones i think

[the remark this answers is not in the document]
caravinci: I think this is as far as we go for config currently:
https://github.com/icgc-argo/dac-ui/blob/3e9649477a2829aff158be7e0f6a41c2c4963e17/package.json#L54

While I haven't made any customisation to my vscode prettier plugin, I wonder if different versions would have different settings 🤔

ciaranschutte (approved):

Issue #175, opened 2021-06-30
Cover plus Sign and Submit Page 


rosibaj:
- [x] Complete cover for pdf
- [x] Complete Sign and Submit page
- [x] missing colons on some vertical table headers, i.e. Applicant info. (as mentioned on https://github.com/icgc-argo/dac-ui/issues/128)
- [x] added lay summary to pdf

anncatton:
@rosibaj I added the `summary` data in this pr

kcullion:
Thanks @anncatton ! This is looking really great. Here's some preliminary style feedback...I will now try testing different content in the application that is now saveable. 

- [ ] we discussed the font changes over slack
- [ ] fix checkbox svgs (will send files over slack)
- [ ] uppercase for DACO
![image](https://user-images.githubusercontent.com/18424138/124967219-2aea2600-dff2-11eb-87a7-86454d57bed2.png)

- [ ] footer: `using` instead of `by`
![image](https://user-images.githubusercontent.com/18424138/124967307-448b6d80-dff2-11eb-9285-b4de4dbc4c00.png)

- [ ]

kcullion:
This feedback will be tested with https://github.com/icgc-argo/dac-ui/issues/128

So this ticket looks good to me @anncatton and @rosibaj

PR #192, opened 2021-07-08
💄 128 - pdf feedback

anncatton:
Feedback fixes for #128 and #175 
Change base font size to `8pt`


components/pages/Applications/PDF/StaticEthics.tsx
```diff
@@ -48,7 +49,14 @@ const ApprovalLetterDocs = ({ data = [] }: { data?: ApprovalDoc[] }) => {
...
 
 const PdfEthicsFormData = ({ data }: { data?: ApplicationData }) => {
-  const ethicsRequired = data?.sections.ethicsLetter.declaredAsRequired;
+  // check if null first, as this means neither has been checked
+  const ethicsRequiredPresent = !isNull(data?.sections.ethicsLetter.declaredAsRequired);
+
+  const ethicsRequiredTrue =
+    ethicsRequiredPresent && data?.sections.ethicsLetter.declaredAsRequired;
+  const ethicsRequiredFalse =
+    ethicsRequiredPresent && !data?.sections.ethicsLetter.declaredAsRequired;
```
ciaranschutte, line 58: @anncatton  I'm inclined to go with something like `areEthicsRequired = ethicsRequiredPresent && data?.sections.ethicsLetter.declaredAsRequired`

Otherwise you're passing `ethicsRequiredFalse` into props requiring booleans, making the logic a little confusing to follow

components/pages/Applications/PDF/StaticEthics.tsx
```diff
@@ -74,15 +82,15 @@ const PdfEthicsFormData = ({ data }: { data?: ApplicationData }) => {
...
       </View>
       <View>
-        <Checkbox TextComponent={OptionTwo} checked={ethicsRequired || true} />
+        <Checkbox TextComponent={OptionTwo} checked={ethicsRequiredTrue as boolean} />
       </View>
       <View style={{ marginTop: '15pt' }}>
-        {ethicsRequired && <StaticAttachLetterMessage isPdf />}
-        {ethicsRequired && (
+        {ethicsRequiredTrue && <StaticAttachLetterMessage isPdf />}
+        {ethicsRequiredTrue && (
```
ciaranschutte, line 93: can probably just do one conditional here by wrapping in a fragment `<></>`

ciaranschutte (changes requested):
two points to have a look at. looks really good

components/pages/Applications/PDF/StaticEthics.tsx
anncatton: I just made the whole `View` conditional.

components/pages/Applications/PDF/StaticEthics.tsx
caravinci: Ditto ☝️ ...
Maybe even simpler: const ethicsRequired = !!data?.sections.ethicsLetter.declaredAsRequired;?

anncatton: the issue here is that i need to differentiate between a `null` value (nothing is checked) and a deliberate `true` or `false`. But i see what @ciaranschutte means by the logic being confusing

components/pages/Applications/PDF/Table.tsx
```diff
@@ -39,7 +39,7 @@ const Table = ({ headers, data }: { headers: { name: string; accessor: string }[
               style={{
                 width: '50%',
                 border: `1pt solid ${defaultTheme.colors.grey_1}`,
-                borderBottom: 0,
+                ...(data.length && { borderBottom: 0 }),
```
caravinci, line 42:
```suggestion
                ...(data.length > 0 && { borderBottom: 0 }),
```

ciaranschutte (approved):

PR #194, opened 2021-07-08
189 - add focus touch to select-one fields

caravinci:
The select fields (e.g. Applicant and Representative `title`) are not saving changes on the first interaction, as the `blur` logic persists only data for fields that had been `touched` before the change.
This logic touches these upon focusing on them, so changes can be saved once the user moves away from them.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -464,6 +464,11 @@ export const useLocalValidation = (
           break;
         }
 
+        case 'focus': {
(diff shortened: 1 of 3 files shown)


components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -464,6 +464,11 @@ export const useLocalValidation = (
           break;
         }
 
+        case 'focus': {
+          ['select-one'].includes(fieldType) && setFieldTouched((prev) => new Set(prev.add(field)));
```
caravinci, line 468: `touch` select-one fields upon focus.

caravinci (commented):
eeehmmm hopefully self documenting? 🧐

anncatton (approved):

PR #195, opened 2021-07-09
💄 181 - pdf download button state

anncatton:
Sets pdf download button text according to application state.
Fixes button appearance when in loading state.
Sets watermark to show on `REVISIONS REQUESTED` as well as `DRAFT` state
Hide Actions buttons when app is in `CLOSED` state


components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -24,10 +25,63 @@ import { getFormattedDate } from '../../Dashboard/Applications/InProgress/helper
...
-const HeaderActions = ({ appId }: { appId: string }): ReactElement => {
+// EXPIRED and RENEWING handling is tbd, CLOSED excludes Action buttons
+const getPdfButtonText = (state: ApplicationState) => {
+  let text = 'PDF';
+
+  if ([ApplicationState.DRAFT, ApplicationState.REVISIONS_REQUESTED].includes(state)) {
+    return (text = 'DRAFT ' + text);
+  }
+  if (isEqual(state, ApplicationState.SIGN_AND_SUBMIT)) {
+    return (text = 'FINALIZED ' + text);
```
ciaranschutte, line 38: might be a little cleaner to do: "return `Finalized ${text}`"

ciaranschutte: and changing the `text` to `const`

components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -24,10 +25,63 @@ import { getFormattedDate } from '../../Dashboard/Applications/InProgress/helper
...
+  }
+  if (
+    [ApplicationState.REVIEW, ApplicationState.APPROVED, ApplicationState.REJECTED].includes(state)
+  ) {
+    return (text = 'SIGNED ' + text);
+  }
+  console.warn('Illegal app state! State: ', state);
```
ciaranschutte, line 45: and return `text` if error. better than undefined.

ciaranschutte (approved):
one stylistic suggestion. looks good !

components/pages/Applications/ApplicationForm/Header/Actions.tsx
anncatton: Yeah I was debating about what to do on this kind of error, but that sounds good for now.

PR #197, opened 2021-07-09
🐛 Rep address can use sameAsApplicant value

anncatton:
Representative address on pdf will use applicant address when `addressSameAsApplicant` is true


caravinci (approved):
👍

Issue #61, opened 2021-05-26
[Data Hookup+UI] Admin Approval flow + Application state


kcullion:
## Goal
To complete the admin approval flow and transition the application to the correct state.

Note, the approval email that gets triggered will be built in the email epic here: https://github.com/icgc-argo/dac-ui/issues/68
(All emails to be built and triggered in this epic: https://github.com/icgc-argo/dac-ui/issues/65)

## Endpoint to use
https://github.com/icgc-argo/dac-api/issues/16

### Confirmation Modal
- [x] build the UI for the approval confirmation modal: When DACO Admin clicks the "Approve" button from https://github.com/icgc-argo/dac-ui/issues/55, they see this modal: https://zpl.io/VkMmDJG that loads in:
a. Application number and institution in brackets
b. Access start date (reporting the current day - i.e. the day this button is clicked)
c. Access end date (reporting one year after the start date)
![image](https://user-images.githubusercontent.com/18424138/120375010-445bcc00-c2e8-11eb-8c15-f57a29cdb1fe.png)

### Application state after approved 
- [x] 1. When confirm approval, the application transitions into the `APPROVED` state 
- [x] 2. Most sections are locked for editing as shown here for section A: https://zpl.io/aBvEw80
- [x] 3. For locked sections, all form fields are disabled and uneditable as seen here for section A: https://zpl.io/aBvEw80
- [x] 4. Section C is unlocked in order to be able to add/remove collaborators

![image](https://user-images.githubusercontent.com/18424138/120373207-11183d80-c2e6-11eb-9aaa-211efdc4aef2.png)

### Application state after approved for DACO Admin
- [x] Keep the orange bar and disable the buttons in this state. 

### Section C 
- [x] 6. Section C table: remove the edit icons from each collaborator table row, the applicant cannot edit already existing collaborator information. But they can still add and remove collaborators. Zeplin: https://zpl.io/2jq8pqp
![image](https://user-images.githubusercontent.com/18424138/120374273-5be68500-c2e7-11eb-8237-b7d7d3952756.png)

### Section E
will be hooked up in this ticket https://github.com/icgc-argo/dac-ui/issues/58

samrichca:
Some work can't be completed right now because the following aren't implemented

- ethics section seeding
- uploading an ethics letter
- choosing whether to upload an ethics letter

[...]

PR #201, opened 2021-07-09
#61: admin approval

samrichca:
- add a modal for approving applications
- after approval:
  - update the collaborators page (with new 'ammendable' state that got added this week)
  - disable the admin buttons on the application page
- add a constant for DATE_FORMAT
- replace internal /applications path with external /applications path


components/pages/Applications/types.ts
```diff
@@ -226,7 +227,6 @@ export type ApplicationsResponseData = {
 };
 
 export enum ApplicationsField {
-  // TODO: nested properties currently can't sort. waiting for backend
```
samrichca, line 229: this got resolved on the backend, removing comment

components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -135,7 +135,7 @@ const HeaderActions = ({
         isLoading={pdfIsLoading}
         onClick={async () => {
           setPdfIsLoading(true);
-          const data = await fetchWithAuth({ url: urlJoin(APPLICATIONS_PATH, appId) })
+          const data = await fetchWithAuth({ url: urlJoin(API.APPLICATIONS, appId) })
```
samrichca, line 138: minor thing, get "/applications" from API not an internal path

components/pages/Applications/ApplicationForm/Header/index.tsx
```diff
@@ -4,13 +4,13 @@ import { css } from '@icgc-argo/uikit';
 import { isEqual } from 'lodash';
 
 import PageHeader from 'components/PageHeader';
+import { DATE_FORMAT } from 'global/constants';
```
samrichca, line 7: moved DATE_FORMAT to a constant

components/pages/Applications/ApplicationForm/Forms/Outline/helpers.ts
```diff
@@ -27,6 +27,7 @@ export const getValidationUIConfig = (status: FormSectionOverallState): Validati
         tagVariant: TAG_VARIANTS.ERROR,
       };
 
+    case FORM_STATES.AMMENDABLE:
     case FORM_STATES.CAN_EDIT:
```
caravinci, line 31: `CAN_EDIT` is `AMMENDABLE`, for the Outline component. I map this states elsewhere, when I'm loading the data from Back End.

I'll add a commit to this PR with that part, as I've already implemented it in my machine.

global/constants/index.ts
```diff
@@ -9,5 +9,7 @@ export const APPLICANT_APPLICATIONS_LABEL = 'My Applications';
 
 export const DATE_RANGE_DISPLAY_FORMAT = 'Y-MM-dd';
 
+export const DATE_FORMAT = 'MMM. dd, yyyy';
```
caravinci, line 12: Thus far, seems this format is only used in the header, so maybe we should name it more specifically, if it's going to be abstracted to a global level.

caravinci (changes requested):
Looking good, but for the Form status handling, which also affects the Outline (table of contents).
I'll open a PR with those changes in a few mins, to ensure things work on that end too.

global/constants/index.ts
samrichca: it's used in modals as well! I'm open to a better name though

caravinci: HMM that does make a difference! good call on abstracting it.
maybe `APPLICATION_DETAILS_DATE_FORMAT` or `APP_DETAILS_DATE_FORMAT`?

samrichca: i'm trying to think of something that clarifies this format has words hmmm

samrichca: it could wind up getting used for an applicant or something so i want to describe the contents of the constant not its use. DATE_TEXT_FORMAT works for me, how bout you?

components/pages/Applications/ApplicationForm/Forms/Outline/helpers.ts
caravinci: For your consideration :) 
https://github.com/icgc-argo/dac-ui/pull/202

caravinci (approved):

samrichca:
tested @caravinci's changes locally, looks good

Issue #62, opened 2021-05-26
[Data Hookup] Request Revisions flow + Application state


kcullion:
## Goal
To complete the request revisions flow and transition the application to the correct state.

Note, the request revisions email that gets triggered will be built in the email epic: https://github.com/icgc-argo/dac-ui/issues/65)

## Endpoint to use
 https://github.com/icgc-argo/dac-api/issues/17 

### Confirmation Modal
This was built here: https://github.com/icgc-argo/dac-ui/issues/63
### Connect this button: 
![image](https://user-images.githubusercontent.com/15824252/120673256-02a26100-c461-11eb-86ca-2fc5b19db212.png)
- if you get a failure, do a retry (visual change from @kcullion !)
- [x] 1. When DACO admin confirms to send the request, the application transitions into the `REVISIONS REQUESTED` state
**Needs endpoint above to be connected to DACO admin button** 
### Application state after requested revisions
- [x] 2. All sections, that do not require revisions, are locked for editing as shown here for section A: https://zpl.io/Vqzx6DZ
- [x] 3. For locked sections, all form fields are disabled and uneditable.
- [x] 4. Unlock the sections that were specified in the request revisions modal, and show an orange pencil edit icon in the table of contents: https://github.com/icgc-argo/dac-ui/issues/63

### Application state after revisions were requested for DACO Admin
- [x] Once DACO Admin sends revisions request, remove the Approve/Request Revisions/Decline buttons from the orange admin bar. 
- [x] The application will be in the same state as it is above for the applicant
![image](https://user-images.githubusercontent.com/18424138/120501643-c6530000-c38f-11eb-9086-a0f7dca5eb0d.png)
** Note the custom expiry should not be here - was left in mocks to do late orange bar should just be blank for now in this state.

kcullion:
This seems to work well @samrichca ! 
Just noticed a couple things and ran them past @blabadi .... made a backend ticket for it
https://github.com/icgc-argo/dac-api/issues/109

kcullion:
@samrichca will that backend ticket affect you at all icgc-argo/dac-api#109? If not, this looks good and I'll move to awaiting release.

samrichca:
that looks like we're not getting the right info from the backend. if we start getting the right info then the icons & access should display as expected

this ticket should be OK but we should test it when he's done, so leave it here as a reminder?

kcullion:
thanks @samrichca I will move this to awaiting release and I added a testing note to the backend ticket to test the UI.

PR #216, opened 2021-07-14
#62: submit request revisions

samrichca:
- enable reviewers submitting revisions (copied this from the approval modal)
- hide admin buttons in revisions requested state

### DIFF
--- a/components/pages/Applications/ApplicationForm/RequestRevisionsBar/index.tsx
+++ b/components/pages/Applications/ApplicationForm/RequestRevisionsBar/index.tsx
@@ -21,6 +21,7 @@ const RequestRevisionsBar = ({ data }: { data: any }) => {
   const { primaryAffiliation } = data.sections.applicant.info;
 
   const buttonsDisabled = ['APPROVED'].includes(state);
+  const buttonsHidden = ['REVISIONS REQUESTED'].includes(state);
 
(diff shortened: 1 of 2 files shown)


components/pages/Applications/ApplicationForm/RequestRevisionsBar/index.tsx
```diff
@@ -75,57 +77,61 @@ const RequestRevisionsBar = ({ data }: { data: any }) => {
...
-                <Icon
-                  css={instructionBoxButtonIconStyle}
-                  fill={theme.colors.white}
-                  height="10px"
-                  name="times"
-                />
-                Reject
-              </span>
-            </Button>
+            {!buttonsHidden && (
```
caravinci, line 80: Think I'd personally opt for making this a positive, to remove "need to think this one" from render stuffs.
Especially since the internal logic for `buttonsHidden` (line 24) itself is easy to reason with even if you invert it.

i.e. `buttonsShown`, so here it'd be `true && <Comp/> || false` rather than `!true & <Comp/> || !false`
does that make sense?

caravinci (approved):
Looking great! 💪 

Made a suggestion comment with a personal consideration for better "cognitive burden" management, but feel free to merge as is if you wish.

PR #259, opened 2021-07-21
add null check when updating local state

caravinci:
because `typeof null === 'object'`
┻━┻ ︵ヽ(`Д´)ﾉ︵ ┻━┻


anncatton (approved):
wheeeeeee

Issue #251, opened 2021-07-21
🐛 Change the date on approve to 2 years in the modal


rosibaj:
![image](https://user-images.githubusercontent.com/15824252/126526858-3ab0b295-11d0-4a73-a381-4feca3a6a700.png)

kcullion:
Tested Jul 22 and I now see the end date is after 2 years. thanks @samrichca 
![image](https://user-images.githubusercontent.com/18424138/126659579-f1806637-d166-4c84-8341-8c9db5c1f0f4.png)

PR #264, opened 2021-07-21
#251: change expiry date from 1 year to 2 years

samrichca:


anncatton (approved):
Breathtaking work @samrichca

PR #266, opened 2021-07-21
Feat/#249 sign feedback

ciaranschutte:
- disable submit and delete when Application in review
- connect delete signed app 
- connect submit
- change some mock data to be dynamic and real. i'm a real boy.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/helpers.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/helpers.tsx
@@ -51,6 +51,8 @@ export const sectionSelector = ({
     validateFieldTouched: (event: any) => void;
   } = useLocalValidation(selectedSection, storedFields, validator(selectedSection));
 
+  const primaryAffiliation = formState.sections.applicant.fields.info_primaryAffiliation.value;
+
   return isLoading || !formState.__seeded ? (
     <Loader />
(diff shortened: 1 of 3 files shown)


samrichca (approved):

components/pages/Applications/ApplicationForm/Forms/helpers.tsx
```diff
@@ -61,6 +63,8 @@ export const sectionSelector = ({
       state={formState.state}
       validateFieldTouched={validateFieldTouched}
       appId={appId}
+      primaryAffiliation={primaryAffiliation}
+      applicationState={formState.state}
```
caravinci, line 67: In https://github.com/icgc-argo/dac-ui/pull/267 I'm passing this as `status`.
I find `state` to be generic in UI/FE terms. Maybe we can rename this as `applicationStatus` or just `status` instead?

ciaranschutte: @caravinci for sure that works! I just popped out for lunch so will do a small PR once im back in

caravinci: No need to worry, Sam insisted we just keep it as is, since it's Thursday... So I've amended my PR to match these changes

PR #272, opened 2021-07-22
comment out Close Application button.

ciaranschutte:
comment out Close Application button
this user flow won't be implemented in time for soft launch

### DIFF
--- a/components/pages/Applications/ApplicationForm/Header/Actions.tsx
+++ b/components/pages/Applications/ApplicationForm/Header/Actions.tsx
@@ -57,6 +57,8 @@ const HeaderActions = ({
       `}
     >
       {/*
+      // Commented out for soft launch. User flow is not implemented yet.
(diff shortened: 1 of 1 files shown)


components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -56,9 +56,10 @@ const HeaderActions = ({
         }
       `}
     >
+      {/*
       <Button onClick={function noRefCheck() {}} size="sm" variant="secondary">
```
anncatton, line 60: if we're not implementing for hard launch, can we put a comment just to document?

ciaranschutte: @anncatton sure can. BAM

anncatton (approved):

PR #273, opened 2021-07-22
reduce font size header bar

ciaranschutte:


components/NavBar.tsx
```diff
@@ -38,6 +38,9 @@ const StyledMenuItem = styled(MenuItem)`
       background-color: ${theme.colors.grey_4};
       color: ${theme.colors.secondary};
     }
+    > span {
+      font-size: 14px !important;
```
anncatton, line 42: sorry to ask but is there any way we can not use `!important` 😀

anncatton (approved):

Issue #262, opened 2021-07-21
*Application Section B: Address same as applicant checkbox.*


rosibaj:
- [ ] **Application Section B: Address same as applicant checkbox.**
If the applicant checks off the "Address is the same as applicant" checkbox in section B, then 
1. Disable the address fields in Section B 
2. If any address fields have values in section A, show those in the disabled fields in section B 
3. If no values are filled out in the address fields in section A, only those need to report the errors for required fields (not the disabled fields in section B)

ui needs to fetch the data from the other sectio in this case

rosibaj:
@kcullion can you add the zeplin for the screen on this ticket?

kcullion:
@samrichca this is the zeplin for this section: https://zpl.io/bz3mDAX
But when the checkbox is checked, the fields just have the regular disabled styling that's already been implemented

kcullion:
tested on QA July 26 and it's working very well, thanks @samrichca !

PR #281, opened 2021-07-23
#262: use same address for applicant & representative

samrichca:


components/pages/Applications/ApplicationForm/Forms/Representative.tsx
```diff
@@ -21,14 +22,20 @@ import StaticRepresentative from '../../PDF/StaticRepresentative';
...
   validateFieldTouched,
 }: {
+  applicantAddress?: FormSectionValidationState_Applicant;
   isSectionDisabled: boolean;
   localState: FormSectionValidationState_Representative;
   validateFieldTouched: FormFieldValidationTriggerFunction;
 }): ReactElement => {
+  const addressSameAsApplicant = !!localState.addressSameAsApplicant?.value;
+  const isAddressDisabled = isSectionDisabled || addressSameAsApplicant;
+  const addressState = addressSameAsApplicant && applicantAddress ? applicantAddress : localState;
```
samrichca, line 37: the `applicantAddress` check is for typescript because it COULD be undefined (it won't be undefined if `addressSameAsApplicant` is true, it'll be an empty object)

caravinci (approved):
🚢

Issue #163, opened 2021-06-28
[Data Hookup] - Show local application errors on Application Load 


rosibaj:
## Goal
A user can save "bad data" as part of form entry.  While the data is saved, the BE keeps the state as incorrect and records the errors that are in that section. 

When an application is loaded, the local validations need to be triggered after load to show the errors in a section to users. 

## Example Data
- https://dac.dev.argo.cancercollaboratory.org/applications/DACO-42
- You can see in the representative body sectoin, i have an `INCOMPLETE` state and `ERROR` in meta
```
        "representative": {
            "address": {
                "building": "",
                "cityAndProvince": "",
                "country": "",
                "postalCode": "",
                "streetAddress": ""
            },
            "addressSameAsApplicant": true,
            "info": {
                "firstName": "Sarumon",
                "googleEmail": "sarumon@example.com",
                "institutionEmail": "@example.com",     <-------------⚠️⚠️⚠️
                "displayName": "Sarumon the White",
                "institutionWebsite": "https://www.facit.on.ca",
                "lastName": "the White",
                "middleName": "",
                "positionTitle": "Warlock",
                "primaryAffiliation": "",
                "suffix": "string",
                "title": "M"
            },
            "meta": {
                "errorsList": [    <-------------⚠️⚠️⚠️ 
                    {
                        "field": "institutionEmail",
                        "message": "Value is not a valid email"
                    },
                    {
                        "field": "primaryAffiliation",
                        "message": "field primaryAffiliation is required"
                    },
                    {
                        "field": "primaryAffililation",
                        "message": "Primary Affiliation must be the same as the Applicant"
                    }
                ],
                "status": "INCOMPLETE"
            }
        },
```

## Expected Outcomes
- when an application ls loaded, all fields should display the contents that are recorded in the BE
- For each field that is listed in the `meta.errorsList` trigger the local validation so that the error shows on that field
- In this example above, this is how the display should look: 
![image](https://user-images.githubusercontent.com/15824252/123656991-64ef4700-d7fe-11eb-93e5-3fe01bca1e32.png)

Note: this does not include the extended error UI for modals (#142) or Primary Affiliation (#54)

caravinci:
As part of this ticket, I'll also ensure the word counter tracks the number upon loading data from BE.
![image.png](https://images.zenhubusercontent.com/5defce3f99bcf0b1b8a98d24/4a911a3a-111e-4e46-b891-d66866f3bad1)

[...]

PR #286, opened 2021-07-26
280 - reduce validation aggro and load persisted failed values

caravinci:
Note: As a reassessment of #163, we won't validate values following the API's `errorsList` array, but all fields that contain a value.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -238,29 +238,26 @@ export const validator: FormSectionValidatorFunction_Main =
           type: 'overall',
         });
 
-        const errorsFromBackEnd = sectionErrorsListFromBackEnd.reduce(
-          (acc, error) => ({
-            ...acc,
-            [error.field]: (acc[error.field] || []).concat(error.message),
-          }),
-          {} as Record<string, any>,
-        );
-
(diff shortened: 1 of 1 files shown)


components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -238,29 +238,26 @@ export const validator: FormSectionValidatorFunction_Main =
...
           async ([field, data]) => {
-            if (
-              Object.keys(errorsFromBackEnd).some((offendingFieldName) =>
-                field.includes(offendingFieldName),
-              )
-            ) {
-              const { error } = await schemaValidator(combinedSchema[origin], {
-                [field]: data.value,
-              });
+            if (data.value) {
```
caravinci, line 243: validate field if it contains a value, rather than relying on the `errorsList`

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -238,29 +238,26 @@ export const validator: FormSectionValidatorFunction_Main =
...
+                    : fieldName,
+                ),
+                data.value,
+              );
 
               const results = {
                 field,
                 section: origin,
                 type: data.type,
+                value: data.value,
```
caravinci, line 260: load all persisted values

caravinci (commented):
Easy PRs get comments too...

anncatton (approved):

Issue #196, opened 2021-07-09
Missing section borders in DAA of application form


anncatton:
A grey border is no longer showing, due to some css inheritance issues on some of Static elements introduced for pdf:

- on DATA ACCESS AGREEMENTS:
![image](https://user-images.githubusercontent.com/18424138/126661251-72cfd776-4be4-4ea0-84f0-3a191ca52ebc.png)

kcullion:
I now see the borders in the DAA section. Thanks @samrichca 
![image](https://user-images.githubusercontent.com/18424138/127348924-6388bed3-0079-4dca-8680-ea514e809180.png)

PR #293, opened 2021-07-27
#196: fix borders on form

samrichca:
- added borders to UI section titles
- removed padding/borders on the sections themselves
- converted H3s to section titles

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/index.tsx
@@ -123,13 +123,6 @@ const ApplicationFormsBase = ({ appId = 'none', setLastUpdated }:
               }
 
               > section {
-                &:not(:first-of-type) {
-                  border-top: 1px solid ${theme.colors.grey_2};
-
-                  &:not(:last-of-type) {
-                    padding-bottom: 25px;
-                  }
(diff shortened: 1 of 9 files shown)


components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -123,13 +123,6 @@ const ApplicationFormsBase = ({ appId = 'none', setLastUpdated }:
               }
 
               > section {
-                &:not(:first-of-type) {
-                  border-top: 1px solid ${theme.colors.grey_2};
-
-                  &:not(:last-of-type) {
-                    padding-bottom: 25px;
-                  }
-                }
```
caravinci, line 132: I'm a bit afraid of this change, as I think some forms rely on that styling for internal components.
I guess QA will catch if this needs to be re-incarnated in some other CSS manner (better + less aggressive than the one being removed)

caravinci (approved):

components/pages/Applications/ApplicationForm/Forms/index.tsx
samrichca: I found one (collaborators) but the others all had section headings 👍

PR #300, opened 2021-07-28
54 - ensure applicant affiliation change triggers validation

caravinci:
I had the wrong check for this trigger.
This change checks for any changes (even for `''`), rather than checking if there's a match.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -337,7 +337,7 @@ export const validator: FormSectionValidatorFunction_Main =
         if (
           validatingApplicant &&
           validatingPrimaryAffiliation &&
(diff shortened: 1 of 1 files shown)


samrichca (approved):

samrichca:
my only note is: `checkMatchingPrimaryAffiliation` is that still being used or is it cruft 👀

caravinci:
> my only note is: `checkMatchingPrimaryAffiliation` is that still being used or is it cruft 👀

Good question! It's still being used to validate the field in other sections (e.g. representative, collaborators) and produce the appropriate error if there's no match.

samrichca:
> > my only note is: `checkMatchingPrimaryAffiliation` is that still being used or is it cruft 👀
> 
> Good question, @samrichca! It's still being used to validate the field in other sections (e.g. representative, collaborators) and produce the appropriate error if there's no match.

cool rock on then!

PR #301, opened 2021-07-28
214 - enable add/edit collaborator button when modal is valid

caravinci:
adds logic to check all required fields in the modal have a value and ensure that the primary affiliation is valid, before enabling the magic button.

Bonus: fixes the `title` field

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
@@ -101,9 +101,9 @@ const Collaborators = ({
         ...dataAcc,
         [prefix]: suffix
           ? {
-            ...dataAcc[prefix],
-            [suffix]: fieldData.value,
-          }
+              ...dataAcc[prefix],
+              [suffix]: fieldData.value,
+            }
           : fieldData.value,
       };
     }, {} as Record<string, any>);
@@ -195,7 +195,12 @@ const Collaborators = ({
     const newModalFields = getInternalFieldSchema(localState.list);
 
     collaboratorCount === newCollaboratorCount || setCollaboratorCount(newCollaboratorCount);
-    setModalHasErrors(Object.values(newModalFields).some((field: any) => field?.error?.length > 0));
+    setModalHasErrors(
+      Object.values(newModalFields).some((field: any) => field?.error?.length > 0) ||
+        !Object.entries(newModalFields)
+          .filter(([fieldName, fieldData]) => fieldName !== 'type' && isRequired(fieldData))
+          .every(([fieldName, fieldData]) => fieldData.value),
+    );
     setModalFields(newModalFields);
   }, [localState]);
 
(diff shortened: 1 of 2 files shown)


components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -195,7 +195,12 @@ const Collaborators = ({
     const newModalFields = getInternalFieldSchema(localState.list);
 
     collaboratorCount === newCollaboratorCount || setCollaboratorCount(newCollaboratorCount);
-    setModalHasErrors(Object.values(newModalFields).some((field: any) => field?.error?.length > 0));
+    setModalHasErrors(
+      Object.values(newModalFields).some((field: any) => field?.error?.length > 0) ||
+        !Object.entries(newModalFields)
+          .filter(([fieldName, fieldData]) => fieldName !== 'type' && isRequired(fieldData))
+          .every(([fieldName, fieldData]) => fieldData.value),
```
caravinci, line 202: hasErrors = `any errors` || `any required values missing`

`type` has a default value, so we don't care about that one.

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -553,7 +554,8 @@ export const useLocalValidation = (
                               [fieldIndex]: {
                                 ...currentField.innerType?.fields[fieldIndex],
                                 ...(value[fieldIndex] || { value }),
-                                error,
+                                // ensure affiliation validation is applied before allowing "save"
+                                error: validatingPrimaryAffiliation ? error || [''] : error,
```
caravinci, line 558: faking an error while a primary affiliation is being filled, to ensure the button only is enabled AFTER the field is validated (which removes this fake error elsewhere)

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -496,7 +501,11 @@ const Collaborators = ({
 
                   <DoubleFieldRow helpText="This must match the applicant’s primary affiliation exactly.">
                     <FormControl
-                      error={!!modalFields.info_primaryAffiliation?.error}
+                      error={
+                        // additional logic to quietly ensure validation is applied before allowing save
+                        !!modalFields.info_primaryAffiliation?.error?.filter((e: string) => e)
+                          .length
```
caravinci, line 507: this avoids highlighting the field as errored (i.e. red outline), if there's only a fake error (which means it hasn't been validated.

TODO: May consider applying this sort of logic as a default for all text input fields, since they're only validated on blur.

caravinci (commented):
short and sweet 🍭

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -397,6 +402,7 @@ const Collaborators = ({
                         disabled={isSectionDisabled}
                         id="list--info_title"
                         onBlur={validateFieldTouched}
+                        onFocus={validateFieldTouched}
```
caravinci, line 405: Last minute addition: fix for the `title` field not working upon first selection.

anncatton (approved):
lgtm

Issue #298, opened 2021-07-28
[UI] Add a blank option to select in title dropdown


kcullion:
## Select/Dropdown enhancement

We need to  include the "Select an option" as an option to select. i.e. to reset the dropdown so it doesn't have anything selected. (this is in section A, B and C modal)

For example, if I were to have specified "Dr." in this field, then wanted to erase that, I have no way of doing so

<img width="459" alt="Screen Shot 2021-06-02 at 11 48 40 AM" src="https://user-images.githubusercontent.com/18424138/120552695-8d358280-c3c5-11eb-960a-8f8c42d83270.png">

So in this screenshot the dropdown options would be: 

-- Select an option --
Dr.
Miss
Mr.
Mrs. 
Ms.
Prof.

rosibaj:
looks great thank you @samrichca !

PR #305, opened 2021-07-29
#298: add default blank option to select options

samrichca:
- changed the first option in select fields to 'select an option' with an empty string value
- empty string is the default for select fields (according to `schemas.ts`)


anncatton (approved):
🌮 🌮 🌮

PR #306, opened 2021-07-29
304 - allow whitespace around text input before validation

caravinci:
Applies `trim` to the yup string schemas, to ensure validating inputs regardless of surrounding whitespace.

Bonus: shows error for required fields filled with blank space.

https://github.com/jquense/yup#stringtrimmessage-string--function-schema


anncatton (approved):
trim all the things!

Issue #58, opened 2021-05-26
Ethics section V3 - POST APPROVAL states (ui + data hookup)


kcullion:
## Goal
This ticket also includes the Ethics section for after the application is approved. 

## Endpoint to use
https://dac.qa.argo.cancercollaboratory.org/api/api-docs/#/Application/post_applications__id__assets_url

## Expected Outcomes after Approval
- [x] 1) If Applicant has selected option 1 (no ethics letter required),  then lock this section shoudl be locked with the grey status in left bar.  No actionns can be taken at this point. 
- [x] 2) If Applicant choses option 2 (required ethics letter) then a few things apply: 
  - this section is unlocked for edits, 
  - option 1 is disabled
  - remove the actions column from the table.  The applicant cannot delete files from this section 
  - the UPLOAD A FILE button should still be enabled. 
![image](https://user-images.githubusercontent.com/15824252/122591573-22c14b00-d031-11eb-96af-8a2f6797e1e1.png)
- [x] 3) Clicking "Upload a file button triggers the browser file upload window to choose a pdf/doc/docx file, with the validations that were already coded. When the user chooses a file and uploads, implement a new confirmation modal for when they upload a new letter. This triggers the email being sent to DACO admin (https://github.com/icgc-argo/dac-ui/issues/73), so we want them to confirm this action. **MOCKUP: https://zpl.io/blOm11X**
![image](https://user-images.githubusercontent.com/18424138/120700686-f24dae80-c47f-11eb-99ce-a4934e46447b.png)

samrichca:
scope note: fixed the ethics letter count & pluralization while working on the table

![image.png](https://images.zenhubusercontent.com/5d2e194240c06427886a21ee/2e0ad65d-db02-4d12-8d9d-855e4bfeb29a)

samrichca:
scope note 2: ethics letters aren't showing up in the table on initial page load or refresh, only when i upload a new letter

rosibaj:
tested Jul 30, 2021 
- was able to upload 2 ethics letters 
- counts are right at the top of the table 
- was able to see 2 letters when clicking away from ethics section and clicking back, and on refresh, 
- saw the correct ethics letters in locked state as well, and on refresh

rosibaj:
thank you @samrichca - everything looks great with the ethics letter addition now!

PR #309, opened 2021-07-29
#58: ethics post approval

samrichca:
- disable radios if section is disabled/locked, or if application is approved
- if user picked option 2, allow upload
  - with a confirmation modal, for this scenario only
- don't allow them to delete letters though
- fix issue where the ethics letters weren't displaying on initial load, and the count was undefined (this was because the ethics letter variable starts out as an object, then becomes an array when it's updated)

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/EthicsLetter/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/EthicsLetter/index.tsx
@@ -19,6 +19,7 @@ import {
 import { isRequired } from '../validations';
 import UploadsTable from './UploadsTable';
 import { getStaticComponents } from 'components/pages/Applications/PDF/common';
+import { ApplicationState } from 'components/pages/Applications/types';
 
 const EthicsLetter = ({
   appId,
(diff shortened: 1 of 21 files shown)


components/pages/Applications/ApplicationForm/Forms/EthicsLetter/index.tsx
```diff
@@ -19,39 +19,49 @@ import {
...
-      // faking event values to keep scope limited
-      target: {
-        id: 'declaredAsRequired',
-        tagName: 'INPUT',
-        type: 'radio',
-        value,
-      },
-      type: 'change',
-    });
+    if (!isSectionDisabled) {
```
caravinci, line 51: This check shouldn't be necessary if the radio group is disabled.
Not a concern in this PR, but is it a possible symptom of a pending `TODO` in the UiKit component itself?

samrichca: it was needed pre-uikit update, removed now, good catch!

caravinci (approved):
🎸 rock on!

PR #328, opened 2021-08-04
Feat/#142 banner modal errors

ciaranschutte:


components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -93,6 +97,12 @@ const Collaborators = ({
   };
 
   const handleCollaboratorCreateOrEdit = useCallback(() => {
+    const collaboratorEmail = modalFields.info_institutionEmail.value;
+
+    if (collaboratorEmail === user?.email) {
+      return setModalBannerError(AddCollaboratorError.CollaboratorIsApplicant);
```
ciaranschutte, line 103: make sure applicant is not adding themselves

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { createRef, ReactElement, useCallback, useEffect, useState } from 'react';
+import React, { createRef, ReactElement, useCallback, useEffect, useState } from 'react';
```
caravinci, line 1: Not needed in React 17+
```suggestion
import { createRef, ReactElement, useCallback, useEffect, useState } from 'react';
```

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
caravinci: I think the intention behind this logic is correct and useful, but we should check instead for the `Applicant` not to be added as a collaborator (exactly like we check for the applicant's Primary affiliation to match.)

Why: The person filling up the form (aka `user`) could be a collaborator, delegated to do so in the name of the PI/Applicant. Furthermore, the `institutional representative` could also be added as a collaborator, so no need to validate for that.

caravinci (changes requested):
Needs a slight adjustment for the validation logic.

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -125,11 +135,18 @@ const Collaborators = ({
...
         }
       })
-      .catch((err: AxiosError) => {
-        if (err?.response?.data?.message) {
-          const responseErrors = JSON.parse(err.response.data.message)?.errors;
-          console.error('Failed to create collaborator.', responseErrors);
+      .catch((err: any) => {
+        const errorCode = err?.error?.response?.data?.code;
+
+        if (errorCode) {
```
ciaranschutte, line 141: add specific handling for http 409 response of collaborator already existing

global/hooks/useAuthContext.tsx
```diff
@@ -112,7 +112,10 @@ export const AuthProvider = ({
         // TODO: log errors somewhere not visible to the user?
         // Leaving this log here pre-release, for troubleshooting
         console.error('Error in fetchWithAuth', { error });
-        throw error;
+        let err: any = new Error(error);
+        err.resp = error.response;
+        console.log('er', err);
+        throw { error };
```
ciaranschutte, line 118: throw entire error object to access http response
other uses of this catch are just logging the error response or using it as a boolean, so it shouldn't have repercussions

components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -63,8 +64,11 @@ const Collaborators = ({
   const [modalVisible, setModalVisible] = useState<'collaborator' | string | null>(null);
   const [modalFields, setModalFields] = useState(getInternalFieldSchema(localState.list));
   const [modalHasErrors, setModalHasErrors] = useState(true);
+  const [modalBannerError, setModalBannerError] = useState<
+    keyof typeof AddCollaboratorError | null
+  >(null);
   const containerRef = createRef<HTMLDivElement>();
-  const { fetchWithAuth, permissions } = useAuthContext();
+  const { fetchWithAuth, permissions, user } = useAuthContext();
```
caravinci, line 71:
```suggestion
  const { fetchWithAuth, permissions } = useAuthContext();
```

caravinci (approved):
Once that `user` import is gone ☝️, we're ready to squash this pup! 🥾

PR #348, opened 2021-08-06
346 - show SOME section status notifications

caravinci:
Fine tuning the validation notifications, we're supposed to show section status changes while the user fills up the form, as long as the status is not "incomplete" (aka error), which is only to be shown if the user moves onto a new section (or leaves the app) before fulfilling the sections' requirements.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -415,10 +415,19 @@ export const validator: FormSectionValidatorFunction_Main =
               },
             }).then(({ data, ...response } = {} as AxiosResponse<any>) => {
               data
-                ? dispatch({
+                ? (!['', SECTION_STATUS.INCOMPLETE, SECTION_STATUS.PRISTINE].includes(
+                    data.sections[origin].meta.status || '',
+                  ) &&
+                    dispatch({
+                      field: 'showOverall',
+                      section: origin,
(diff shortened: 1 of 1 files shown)


components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -415,10 +415,19 @@ export const validator: FormSectionValidatorFunction_Main =
               },
             }).then(({ data, ...response } = {} as AxiosResponse<any>) => {
               data
-                ? dispatch({
+                ? (!['', SECTION_STATUS.INCOMPLETE, SECTION_STATUS.PRISTINE].includes(
+                    data.sections[origin].meta.status || '',
+                  ) &&
```
caravinci, line 420: check whether the updated section status is any other than incomplete (or pristine, which would be an API error after a patch, for good measure.)

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -415,10 +415,19 @@ export const validator: FormSectionValidatorFunction_Main =
...
+                    dispatch({
+                      field: 'showOverall',
+                      section: origin,
+                      type: 'sectionOverall',
+                      value: true,
```
caravinci, line 425: and if so, dispatch the order to show the icon on the application outline.

anncatton (approved):

PR #365, opened 2021-08-11
329 - validate Applicant emails at collaborator modal

caravinci:
Abstracts the comparison of fields with the Applicants' fields into a reusable helper, in case we have to add other checks anywhere.

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -6,7 +6,7 @@ import { AuthAPIFetchFunction } from 'components/pages/Applications/types';
 import { API } from 'global/constants';
 import { useAuthContext } from 'global/hooks';
 
-import { sectionsOrder, sectionStatusMapping } from '../constants';
+import { applicantFieldNames, sectionsOrder, sectionStatusMapping } from '../constants';
 import {
   FormFieldType,
   FormSectionNames,
@@ -24,7 +24,7 @@ import {
 import {
   getFieldDataFromEvent,
   getValueByFieldTypeToPublish,
-  checkMatchingPrimaryAffiliation,
+  checkMatchingApplicant,
   schemaValidator,
   sectionFieldsSeeder,
   getValueByFieldTypeToValidate,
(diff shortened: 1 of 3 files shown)


components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -290,8 +287,7 @@ export const validator: FormSectionValidatorFunction_Main =
             ...(error
               ? { error }
               : !validatingApplicant &&
-                validatingPrimaryAffiliation &&
-                checkMatchingPrimaryAffiliation(data.value, applicantPrimaryAffiliation)),
+                checkMatchingApplicant(origin, field, data.value, applicantData)),
```
caravinci, line 290: using that improved helper for every field in every section now, in case there's anything else we should be checking for based on the applicant.

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -347,12 +341,14 @@ export const validator: FormSectionValidatorFunction_Main =
...
         if (
           validatingApplicant &&
-          validatingPrimaryAffiliation &&
-          applicantPrimaryAffiliation !== value
+          ((field.includes(applicantFieldNames.AFFILIATION) &&
+            value !== applicantData[applicantFieldNames.AFFILIATION]) ||
+            (field.includes(applicantFieldNames.EMAIL) &&
+              value !== applicantData[applicantFieldNames.EMAIL]) ||
+            (field.includes(applicantFieldNames.GMAIL) &&
+              value !== applicantData[applicantFieldNames.GMAIL]))
```
caravinci, line 351: this right here just makes sure the representative does an overall validation again if the affiliation or emails change.

This and the helper could be improved based on whether we add more special fields.

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -645,7 +639,8 @@ export const useLocalValidation = (
           break;
         }
 
-        case 'change': {
+        case 'change':
+        case 'mousedown': {
```
caravinci, line 643: this line right here is why the modals were malfunctioning. oops!

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -661,21 +656,18 @@ export const useLocalValidation = (
             changes && updateLocalState(changes);
           } else if (fieldType.includes('Modal')) {
             fieldValidator(field, value);
-          } else {
-            const shouldPersistData = ['checkbox', 'radio', 'select-one'].includes(fieldType);
-            const checkMultiSelectValue =
-              fieldType === 'select-one' && Array.isArray(value) ? value[0] : value;
+          } else if (fieldType !== 'select-one') {
```
caravinci, line 659: extra bit here to avoid making duplicate requests from the `country` fields

caravinci (commented):
Please let me know how I can make these more helpful...

anncatton (approved):

Issue #176, opened 2021-06-30
Redirect/Login


rosibaj:
When you are taken to a URL and arent logged in
- direct the user to login 
- after login the user should be taken to the original page

samrichca:
![image](https://user-images.githubusercontent.com/3540896/125863760-b593b974-a7dd-47ca-b926-a16c196615ae.png)

![image](https://user-images.githubusercontent.com/3540896/125863790-208603cd-a08e-442a-8dfa-8e7a7742781d.png)


Debugging note: In the browser console turn on `preserve logs` or you'll never see some of the console logs.

samrichca:
the branch is feat/login-redirect

pausing this for now to decide how to proceed with loading

PR #374, opened 2021-08-16
[DRAFT] #176: login redirects

samrichca:
login redirects:
- if a user tries to access a protected route or API without a valid JWT, they'll be logged out to the homepage with a `redirect` query added to the URL. when they login, they'll go back to where they were.
- this will also apply if a user is on a public page and logs in. they'll be redirected to the page they were on.

other things:
- add `usePageContext` from platform-ui so we can see path & query in child components
- abstracting some stuff to separate functions & files
- use `<Loader />` component in `/logged-in` & homepage


ciaranschutte:
On login the redirect works but I get a couple of flashes of other pages before that.
Ping me and I can share video.

samrichca:
> On login the redirect works but I get a couple of flashes of other pages before that.
> Ping me and I can share video.

Yeah I noticed that too. We'd have to assume *any* user is potentially in the redirect flow and show a loader as the default. We could do that with the global loader from platform. I tried it here and it worked but required some more thought about how we load each page in different scenarios. So I've proposed it as a ticket for the next sprint.

samrichca:
closing because we're going in a different direction

PR #375, opened 2021-08-16
Feat/#103

ciaranschutte:
direct user to error page for failed routes


components/Root.tsx
```diff
@@ -40,55 +40,47 @@ export const ModalPortal = ({ children }: { children: React.ReactElement }) => {
   const mounted = useMounted();
   return ref
     ? ReactDOM.createPortal(
-      <div
```
ciaranschutte, line 43: prettier ping pong

components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -29,16 +30,15 @@ const ApplicationForm = ({ appId = 'none', isAdmin = false }): ReactElement => {
         setData(res.data);
       })
       .catch((err: AxiosError) => {
-        setError(err);
+        console.error('Application form error', err);
+        return router.push('/_error');
```
ciaranschutte, line 34: remove local error state. route to error page on error

components/pages/Error/Maintenance.tsx
```diff
@@ -23,59 +23,66 @@ import { Row, Col } from 'react-grid-system';
 import Typography from '@icgc-argo/uikit/Typography';
 import Image from 'next/image';
 import { css } from '@emotion/core';
+import { PageContainer } from '@icgc-argo/uikit/PageLayout';
```
ciaranschutte, line 26: new maintenance page

components/pages/Applications/ApplicationForm/index.tsx
samrichca: what's the difference between `/_error` and `/error`? also pls add these to `internalPaths`

components/pages/Error/index.tsx
```diff
@@ -24,32 +24,25 @@ import Container from '@icgc-argo/uikit/Container';
...
-          `}
-        >
-          {children}
-        </Container>
-      </div>
-      <Footer />
-    </PageContainer>
-  );
-};
+export default ({ children }: { children: React.ReactNode; hideNavbarLinks?: boolean }) => (
```
samrichca, line 27: i get this warning in terminal

```sh
warn  - ./components/pages/Error/index.tsx
Anonymous arrow functions cause Fast Refresh to not preserve local component state.
Please add a name to your function, for example:

Before
export default () => <div />;

After
const Named = () => <div />;
export default Named;
```

components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -10,10 +10,11 @@ import ContentError from 'components/placeholders/ContentError';
 import ApplicationHeader from './Header';
 import ApplicationFormsBase from './Forms';
 import RequestRevisionsBar from './RequestRevisionsBar';
+import GenericError from 'components/pages/Error/Generic';
```
samrichca, line 13: unused imports for GenericError & ContentError in a few places

components/pages/Applications/ApplicationForm/index.tsx
ciaranschutte: `_error` is a next catch all error route. but we can also manually nav to it.

components/Root.tsx
anncatton: yeah we gotta find out where everyone is getting their config from

components/pages/Error/index.tsx
anncatton: tbh i think it is better to name components explicitly so that they are more identifiable in dev tools. Feels like anonymous functions don't provide much benefit.

anncatton (approved):
Just a small comment, otherwise 🌮 🌮 🌮

PR #383, opened 2021-08-18
DRAFT jwt refresh token

samrichca:
- remove auth from `_app.tsx`
  - this was useful in platform because we could handle auth cookies in `getInitialProps`, but we're using localStorage in daco


components/pages/Error/index.tsx
```diff
@@ -24,7 +24,7 @@ import Container from '@icgc-argo/uikit/Container';
 import { css } from '@emotion/core';
 import { PageContainer } from '@icgc-argo/uikit/PageLayout';
 
-export default ({
+const Error = ({
```
samrichca, line 27: fixing a next error

components/Root.tsx
```diff
@@ -131,10 +129,9 @@ const Root = ({
       `}
       </style>
       <Head />
-      <AuthProvider egoJwt={egoJwt}>
```
samrichca, line 134: stop passing down `egoJwt` from `_app` to `Root` to `AuthProvider`

this was initially added because it's how we do things in platform. in platform we use cookies and access them in `getInitialProps` in `_app`, then pass the JWT down to `<AuthProvider>`. in daco we use localStorage and can't access the JWT until after the page renders. so checking auth in both _app and AuthProvider is redundant, and they become out of sync with each other.

global/hooks/useAuthContext.tsx
```diff
@@ -74,22 +96,52 @@ export const AuthProvider = ({
...
-      setTokenState('');
+    } else if (isValidJwt(egoJwt)) {
+      setTokenState(egoJwt);
     }
-  } else if (isValidJwt(egoJwt)) {
-    setTokenState(egoJwt);
   }
 
-  // TODO: decide if we want these for all types of requests or only POST
-  axios.defaults.headers.post['Content-Type'] = 'application/json;charset=utf-8';
```
samrichca, line 90: created an axios instance instead of using these defaults

global/utils/refreshJwt.ts
```diff
@@ -0,0 +1,35 @@
+import urlJoin from 'url-join';
```
samrichca, line 1: this is straight from platform. updated some names/imports for this repo

global/hooks/useAuthContext.tsx
```diff
@@ -52,19 +73,20 @@ const AuthContext = createContext<T_AuthContext>({
...
 }: {
   children: React.ReactElement;
-  egoJwt: string;
 }) => {
   // TODO: typing this state as `string` causes a compiler error. the same setup exists in argo but does not cause
   // a type issue. using `any` for now
+  const egoJwt = typeof window !== 'undefined' && localStorage.getItem(EGO_JWT_KEY) || '';
   const [isLoading, setLoading] = useState<boolean>(true);
   const [token, setTokenState] = useState<string>(egoJwt);
+  const [isLoadingRefreshToken, setLoadingRefreshToken] = useState<boolean>(false);
```
samrichca, line 84: AuthContext runs often, so we need to suspend auth checks while the refresh token is loading

global/hooks/useAuthContext.tsx
```diff
@@ -40,6 +44,23 @@ type T_AuthContext = {
...
+const axiosInstance = axios.create({
+  headers: {
+    post: {
+      ['Content-Type']: 'application/json;charset=utf-8',
+      ['Access-Control-Allow-Origin']: '*'
+    }
+  }
+});
+
+var maxConcurrent = 1;
```
samrichca, line 56: not sure if I need promise queues if I added a state for loading refresh tokens (see below). will test removing this

pages/_app.tsx
```diff
@@ -37,27 +33,9 @@ const App = ({
   pageProps: PageConfigProps;
   ctx: NextPageContext;
 }) => {
-  const [initialJwt, setInitialJwt] = useState<string>('');
   const { NEXT_PUBLIC_MAINTENANCE_MODE_ON } = getConfig();
-
-  useEffect(() => {
```
samrichca, line 43: removing auth from _app because it's redundant

it was built this way to match platform. in platform we check cookies in getInitialProps and seed AuthContext with that info. since we're using localStorage here rather than cookies, we can't check auth in getInitialProps, so checking in _app is redundant.

samrichca:
closing because we're going in a different direction

PR #389, opened 2021-08-23
flash data successful submission banner

ciaranschutte:
- add flash data, with reset on route change
- avoids adding more local state
- avoids ugly navigable urls eg. user navigates to `.../submissionSuccess=true`  with zero context


anncatton (approved):
lgtm 🌮 🍾

PR #407, opened 2021-09-01
Feat/ts husky check

ciaranschutte:
- fixes build
- adds Husky with `pre-push` hook of type-checking to ensure this is flagged before reaching dev again
- adding `npm run build` is a little slow right now but we can change it to this if we see fit


caravinci (approved):
LGTM 👍

PR #422, opened 2021-09-15
Feat/#369 close app pre approval

ciaranschutte:
- add pre approval closed flow
- move entire form state up a level, so we can use `__refetchAllData` function in the top bar header too
- top progress bar always in closed state

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/types.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/types.ts
@@ -254,7 +254,8 @@ export type FormValidationAction = {
   value?: any;
 };
 
-interface FormValidationState_Base {
+export type RefetchData = (action?: Partial<FormValidationAction>) => void;
+export interface FormValidationState_Base {
   appId: string;
   approvedAtUtc?: string;
   approvedBy?: string;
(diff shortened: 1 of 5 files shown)


components/pages/Applications/ApplicationForm/Forms/types.ts
```diff
@@ -254,7 +254,8 @@ export type FormValidationAction = {
   value?: any;
 };
 
-interface FormValidationState_Base {
+export type RefetchData = (action?: Partial<FormValidationAction>) => void;
```
caravinci, line 257: Perhaps I'm nitpicking, yet for the sake of uniformity, methinks this should be called RefetchDataFunction or something like that.
Thoughts?

components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -52,17 +55,23 @@ const ApplicationForm = ({ appId = 'none', isAdmin = false }): ReactElement => {
...
       });
   }, [lastUpdated]);
 
-  return isLoading || (data && Object.values(data).length < 1) ? (
+  return isAppLoading || (data && Object.values(data).length < 1) ? (
     <Loader />
   ) : (
     <>
-      <ApplicationHeader data={data} />
+      <ApplicationHeader refetchAllData={formState.__refetchAllData} data={data} />
```
caravinci, line 66: Review note: ☝️ this right here is the gist of relocating `useFormValidation`.

Good call, @ciaranschutte. I had meant to elevate the refetching to this level.

components/pages/Applications/ApplicationForm/Forms/types.ts
ciaranschutte: sounds great!

caravinci (approved):
🥾 squash this puppy!

PR #423, opened 2021-09-16
allow admin to add/remove collaborators

blabadi:


components/pages/Applications/ApplicationForm/Forms/Collaborators/index.tsx
```diff
@@ -82,12 +82,9 @@ const Collaborators = ({
     keyof typeof AddCollaboratorError | null
   >(null);
   const containerRef = createRef<HTMLDivElement>();
-  const { fetchWithAuth, permissions } = useAuthContext();
+  const { fetchWithAuth } = useAuthContext();
   const theme = useTheme();
-
-  const isAdmin = permissions.length > 0 && isDacoAdmin(permissions);
   const disableActions =
-    (isAdmin && applicationState === ApplicationState.APPROVED) ||
```
blabadi, line 90: we no longer need to check if the user is an admin, we rely on the section state from the backend same as the normal user flow

ciaranschutte (approved):

PR #424, opened 2021-09-16
add reject modal

samrichca:
- add rejection modal (copy pasted from approval modal tbh)
- add button to open the modal


ciaranschutte (approved):
this looks good!
i think grabbing the text value from the text area would be better than state... i've noticed some delays on the uikit inputs.
but i don't think there's a straightforward way to get the value from this because of how it's wrapped :(

PR #461, opened 2021-09-27
Feat/#414 toasts

ciaranschutte:


ciaranschutte:
@samrichca moved enum out to be an import
@anncatton reverted the messsage popup i deleted

PR #543, opened 2022-03-03
💄💬 218 - data agreement updates

anncatton:
Updates DAA and terms
Fixes issue where watermark was appearing only on every section, not every page
Refactors some external URLs
Updates .env.schema file
Adds daco email to config - need to add in helm values file


components/pages/Applications/PDF/StaticDataAccessAgreement.tsx
```diff
@@ -381,17 +348,20 @@ const StaticDataAccessAgreement = ({
...
+          {termsAndConditionsList(ListComponent, LinkComponent, TextComponent).map(
+            (TermElement, i) => (
+              <TermElement count={i + 1} key={i} />
+            ),
+          )}
         </ListAsTypography>
 
         <TextComponent
+          style={{ paddingTop: '10pt' }}
           css={css`
```
ciaranschutte, line 360: @anncatton can we use css prop for the style too?

anncatton: Not for the pdf components, they need the `style` prop

ciaranschutte: aha!

ciaranschutte (approved):

samrichca (approved):

PR #552, opened 2022-03-09
Maintenance

ciaranschutte:
- adds export stuff
- makes maintenance public to make export easier
- removes Nextjs Image component usage as `next export` doesnt work with it
  - can user different loader if we want, right now just replaced with static img


anncatton (approved):
thanks @ciaranschutte !!

Issue #538, opened 2022-02-24
JWT Refresh - Fixes for MVP release


kcullion:
## Goal
To implement the fixes necessary for DACO to avoid logging a user out after 30 min.

samrichca:
On page load & DAC-API calls, if there's a token & it's invalid, make a request to the refresh endpoint.

PR #553, opened 2022-03-09
✨ #538: Add refresh tokens

samrichca:
- Refresh the user's JWT on page change and when fetching from DACO-API.


global/utils/auth/refreshJwt.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { isValidJwt } from '../egoTokenUtils';
+
+const { NEXT_PUBLIC_EGO_API_ROOT, NEXT_PUBLIC_EGO_CLIENT_ID } = getConfig();
+const egoRefreshUrl = urlJoin(
+  NEXT_PUBLIC_EGO_API_ROOT,
+  `/oauth/refresh?client_id=${NEXT_PUBLIC_EGO_CLIENT_ID}`,
+);
+
+var maxConcurrent = 1;
+var maxQueue = Infinity;
```
ciaranschutte, line 33: I haven't looked too closely at promise-queue module, but do we need these on global scope?

samrichca: That is a really good question, it works this way on platform-UI so I didn't question it. I'll look into it.

anncatton: i think that keeps us from having several instances of queue?

ciaranschutte: alternative is a singleton I guess? It's a pretty old module. most of the commits past 2015 are just dependency updates and versioning so wondering if var usage is needed or just older lib.  if it works, it works :)

pages/_app.tsx
```diff
@@ -64,22 +65,43 @@ const App = ({
...
-      // redirect to logout when token is expired/missing only if user is on a non-public page
-      if (!Component.isPublic) {
-        Router.push({
-          pathname: '/',
-          query: { session_expired: true },
-        });
+    const handleAuth = async () => {
+      console.log('APP handleAuth');
+      const egoJwt = localStorage.getItem(EGO_JWT_KEY) || '';
+      if (isValidJwt(egoJwt)) {
```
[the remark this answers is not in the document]
anncatton: sorry my indentation didn't translate here 😕

pages/_app.tsx
```diff
@@ -64,22 +65,43 @@ const App = ({
     };
   }, []);
 
+  const logout = () => {
```
blabadi, line 68: should the logout call DELETE /oauth/refresh ?

anncatton: ah good catch!

pages/_app.tsx
```diff
@@ -64,22 +65,43 @@ const App = ({
...
+    setInitialJwt('');
+    localStorage.removeItem(EGO_JWT_KEY);
+    // redirect to logout when token is expired/missing only if user is on a non-public page
+    if (!Component.isPublic) {
+      Router.push({
+        pathname: '/',
+        query: { session_expired: true },
```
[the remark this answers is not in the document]
samrichca: Good idea! Removed it in logout() in AuthContext, which is called by the nav bar logout button.

pages/_app.tsx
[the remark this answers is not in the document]
samrichca: I understood what you meant though - updated that part.

pages/_app.tsx
samrichca: I added a delete call, and I'm deleting the localStorage JWT after the call is complete.

blabadi: looks good thanks !

Issue #540, opened 2022-02-25
Build a static page for daco.icgc.org to redirect to


kcullion:
OICR IT to implement a static html page that will be posted on daco.icgc.org for public users to read before they click to redirect to the new daco.icgc-argo.org.

Content is not confirmed yet, but we can build this page for now: https://zpl.io/5EkPZE8

![image](https://user-images.githubusercontent.com/18424138/156597335-cb0c2129-ad28-4d36-be7b-0079111e0e5e.png)

Page to live on daco.icgc-argo.org/new-website-notice

kcullion:
@anncatton just a quick picky update from me - just to add more space in between the "chunks of headline/paragraph"

![image](https://user-images.githubusercontent.com/18424138/158241474-d0fe053f-68ae-4788-83c3-5f88f51240b9.png)

akachru-github:
@kcullion to ping Amber on updating content.

Issue #539, opened 2022-02-24
Implement a maintenance page for release


kcullion:
Implement this page, but within the DACO website:
https://platform-ui.qa.argo.cancercollaboratory.org/maintenance

- [x] Hide the top "apply for access" and login buttons
- [x] Hide the footer links to platform.icgc-argo.org

![image](https://user-images.githubusercontent.com/18424138/155809170-51341664-dfc4-4693-946a-ec7e9a7d489f.png)

akachru-github:
Let's confirm how platform does this. 
Confirm with IT how to host this in production. Deadman switch would be nice to have, but minimally we would require ability to manually switch over to this from the LB.

akachru-github:
IT ticket: https://servicedesk.oicr.on.ca/servicedesk/customer/portal/2/IT-40759

kcullion:
The netlify page looks good to me! I'll leave this in QA for Atul to review Ann's question

Issue #218, opened 2021-07-15
[ContentUpdates] Data Access Agreement


rosibaj:
Kim emailed Ann the agreement.

## Introduction
- [x] As confirmed on Feb 28 meeting, the first intro section goes above the agreement.
**TODO - ask Amber to review this again because we did put a lot of that copy within the first Application Terms section. and some of it is written like agreement checkboxes**

![image](https://user-images.githubusercontent.com/18424138/156448556-3a34ba58-093b-4d85-8c1c-14b7a58dc53f.png)
![image](https://user-images.githubusercontent.com/18424138/156448393-6f4db8cb-29b1-498f-b62d-892290bea333.png)

- [x] the second link in this copy is broken in the word doc and should open this in a new tab: https://www.icgc-argo.org/page/136/e83-data-access-compliance-office-policies-and-procedures
![image](https://user-images.githubusercontent.com/18424138/156451333-40678fa0-9e25-435c-a02b-c67a9f297387.png)

## Definition and Agreements
- [x] The rest of the definitions and agreements can be updated using the word doc
![image](https://user-images.githubusercontent.com/18424138/156449854-2c849d93-fc0e-43b5-a248-76581766a78a.png)

- [x] Use the links in the word doc if they work. I confirmed these ones with Amber:
- [x] The definition references this link under ICGC Controlled Data item: **E8.1 Data Access Framework of the Consortium’s Policies and Guidelines**  should open in a new tab: https://www.icgc-argo.org/page/134/e81-data-access-framework 
- [x] In section 21 of the agreement there’s a link to the **Data Breach Policy** should open this in a new tab: https://www.icgc-argo.org/page/137/e84-data-breach-policy 
- [x] At the end of the terms, change Data Access Agreement last updated on: June 12, 2021
to `Data Access Agreement Version 2.0 Published: February 2022`

## Agreeements
As confirmed on Feb 28 meeting, nothing needs to change here. Keep all of the IT agreements for Amber and Judith to review. 
![image](https://user-images.githubusercontent.com/18424138/156453504-202bbe0b-0f0e-4407-9689-483db627fa8c.png)

kcullion:
This looks good for DACO team's review...just waiting on an answer about the weird hyphenation, so I'll leave this here

kcullion:
@anncatton you just missed one of the weird hypenations: 
[...]

PR #562, opened 2022-03-14
🔖 Rc/0.39.0

anncatton:
#218 
#540 
#539


package.json
```diff
@@ -1,13 +1,15 @@
...
   "private": true,
   "scripts": {
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
+    "export": "next export",
     "type-check": "tsc -p ./tsconfig.json",
-    "prettier-staged": "pretty-quick --staged"
+    "prettier-staged": "pretty-quick --staged",
+    "export-maintenance": "rm -rf ./out && npm run build && EXPORT_PATH=maintenance NEXT_PUBLIC_MAINTENANCE_MODE_ON=true next export && rm ./out/maintenance/404.html && rm -r ./out/_next"
```
ciaranschutte, line 12: just noticed  - I think this will early exit if there is no `./out` folder already @anncatton  everything else looks ok

ciaranschutte (approved):

package.json
anncatton: kk...i'll test out locally but will make this release in the meantime!

PR #565, opened 2022-03-22
#538 - reload page after logout

samrichca:
Reload the page after logout because the page wasn't updating after logout on develop.

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -29,9 +29,9 @@ import {
   SUBMISSION_SUCCESS_CHECK,
 } from 'global/constants';
 import { isValidJwt } from 'global/utils/egoTokenUtils';
-import Router, { useRouter } from 'next/router';
+import { useRouter } from 'next/router';
 import refreshJwt from 'global/utils/auth/refreshJwt';
-import deleteTokens from 'global/utils/auth/deleteTokens';
+import logoutUtil from 'global/utils/auth/logoutUtil';
 import NewWebsiteNotice from 'components/pages/NewWebsiteNotice';
 
 const resetFlashData = () => {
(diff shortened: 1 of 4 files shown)


ciaranschutte:
why are there two logout functions? 
There seems to be some parts of duplicated logic between auth hook and _app which maybe going back in the direction we came from. Can we constrain everything to the auth hook? Some minor tweaking is needed of course but probably better to do while we're early in this. eg. pass a path to the auth hook logout to account for a custom path

samrichca:
> why are there two logout functions? There seems to be some parts of duplicated logic between auth hook and _app which maybe going back in the direction we came from. Can we constrain everything to the auth hook? Some minor tweaking is needed of course but probably better to do while we're early in this. eg. pass a path to the auth hook logout to account for a custom path

(Discussed on Slack) I have to share auth logic between app.tsx and the auth hook because I wasn't able to put all of the auth logic into the auth hook, and get a working solution, in the timeframe required. This is good feedback though and something to keep in mind.

I've revised my PR to make a utility for logging out which is shared by app.tsx & the auth hook.

pages/_app.tsx
```diff
@@ -106,7 +98,7 @@ const App = ({
         logout();
       }
     };
-    handleAuth();
+    if (!ctx.query?.session_expired) handleAuth();
```
anncatton, line 101: just a code style quibble...can `handleAuth()` be put inside a `{}` for the if statement? i always find this one line style harder to read

anncatton (approved):

pages/_app.tsx
samrichca: Done!

PR #566, opened 2022-03-23
DRAFT #538 - refresh tokens

samrichca:
# Revert Jenkinsfile before merging

This PR contains all the changes for refresh tokens. Some were initially merged and reverted, but are present on this branch.

- Add NextJS `Link` component to local links.
- Change `logout` function to selectively redirect the user or add `session_expired` URL param. It will also now delete the refresh token.
- Restructure `Root` component so `AuthContext` has access to `PageContext` and `ThemeProvider`.
- Move `fetchWithAuth` to useDataContext (new file).
  - Add refresh JWT logic to `fetchWithAuth`.
- Split up `tokenLoading` into `userLoading` and `dataLoading`.
- Add `egoRefreshUrl` to external paths.
- Add `homepage_path` to internal paths.
- Remove auth logic from `app.tsx` and `/logged-in`.
- Use queues for auth and fetch requests.
- For links to an application, add `section=terms` to avoid a redirect.
- Rewrite AuthContextProvider
  - Improve loading condition, and block the entire page if auth is loading.
  - Add router events for checking auth on root changes. 
  - Add a useEffect hook for checking auth on initial page load and hard refreshes.
  - Add `getUserJwt` to get the latest JWT, refresh the JWT, or log the user out.
  - Add `forceLogout` to logout expired users in a consistent way.
  - Move fetch to its own hook (useDataContext).


global/utils/auth/deleteTokens.tsx
```diff
@@ -26,7 +26,7 @@ const deleteTokens = () => {
   console.log('DELETE TOKENS jwt in localStorage', storedToken.slice(-10));
 
   if (storedToken) {
-    fetch(egoRefreshUrl, {
+    return fetch(egoRefreshUrl, {
```
samrichca, line 29: Returns a Promise so I can use `.then()`

global/utils/auth/logoutUtil.tsx
```diff
@@ -20,10 +20,20 @@
 import deleteTokens from './deleteTokens';
 import Router from 'next/router';
 
-const logoutUtil = ({ isManual = false, isPublic = false } = {}) => {
+const logoutUtil = ({
+  handleTokenState = () => {},
+  isManual = false,
+  isPublic = false,
+}: {
+  handleTokenState: ((jwt: string) => void | PromiseLike<void>) | undefined;
```
samrichca, line 28: This was the suggested type for this prop/argument. Open to suggestions!

global/hooks/useAuthContext.tsx
```diff
@@ -67,19 +69,26 @@ export const AuthProvider = ({
...
     console.log('LOGOUT manual?', isManual);
-    setTokenState('');
-    logoutUtil({ isManual });
+    logoutUtil({ handleTokenState, isManual });
   };
 
-  if (token) {
-    if (isValidJwt(token) && !egoJwt) {
-      setTokenState('');
+  useEffect(() => {
```
samrichca, line 83: Had to put this in a `useEffect` because I got a React error about trying to update state in App.tsx before the app has finished rendering.

global/hooks/useAuthContext.tsx
```diff
@@ -67,19 +69,26 @@ export const AuthProvider = ({
   const { NEXT_PUBLIC_DAC_API_ROOT } = getConfig();
   const toaster = useToaster();
 
+  const handleTokenState = (jwt: string) => {
```
samrichca, line 72: Updates app.tsx state and AuthContextProvider state at the same time.

global/utils/auth/logoutUtil.tsx
```diff
@@ -33,10 +43,6 @@ const logoutUtil = ({ isManual = false, isPublic = false } = {}) => {
         : {
             query: { session_expired: true },
           }),
-    }).then(() => {
```
samrichca, line 36: I think this reload is redundant.

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,160 @@
...
-      } else {
-        console.log('FETCH localStorage token is not valid');
-        const refreshedJwt = (await refreshJwt().catch(logout)) as string;
-        if (isValidJwt(refreshedJwt)) {
-          console.log('FETCH refreshed token is valid');
-          setTokenState(refreshedJwt);
-          fetchToken = refreshedJwt;
+  useEffect(() => {
+    // on page change - doesn't detect navigating to
+    // different sections of an application
```
ciaranschutte, line 98: I think the comments on these two useEffects are confusing.
the code is not really tied to page change, it's tied to any related state.

you could change it to a description of what the useEffect is for.
something like: "check if on logged-in page when mounting AuthProvider" and for the other "check page route on every render"

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,160 @@
...
+        if (isValidJwt(storedJwt)) {
+          console.log('🎃 USER valid localStorage token', storedJwt.slice(-10));
+          setUserJwtState(storedJwt);
         } else {
-          console.log('FETCH refreshed token is not valid');
-          logout();
-          setLoading(false);
-          return Promise.reject(undefined);
+          console.log('🎃 USER non valid localStorage token', storedJwt.slice(-10));
+          const refreshedJwt = (await refreshJwt().catch(logout)) || '';
```
ciaranschutte, line 135: what does the || catch here?
Either we get a refreshed token value or we logout, and logout presumably cleans up auth state.

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,160 @@
...
+};
+
+const AuthContext = createContext<T_AuthContext>(authContextDefaultValues);
+
+export const AuthProvider = ({ children }: { children: ReactElement }) => {
+  const [userJwtState, setUserJwtState] = useState<T_AuthContext['token']>(
+    authContextDefaultValues.token,
+  );
+  const [user, setUser] = useState<T_AuthContext['user']>(authContextDefaultValues.user);
+  const [permissions, setPermissions] = useState<T_AuthContext['permissions']>(
```
ciaranschutte, line 62: I don't think there's a need to have `user` and `permissions` in state.
you can derive both from the jwt and still return it in context.

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,160 @@
...
-      });
-  };
-
-  const userInfo = token ? decodeToken(token) : null;
-  const user = userInfo ? extractUser(userInfo) : undefined;
-  const permissions = getPermissionsFromToken(token);
-
-  isLoading && token && user && setLoading(false);
+  useEffect(() => {
+    // populate user state when userJwt updates & initial render
```
ciaranschutte, line 159: I like this comment. This useEffect may not be needed if you remove `user` and `permissions` from state.

global/hooks/useDataContext.tsx
```diff
@@ -0,0 +1,140 @@
...
+          handleUserJwt(refreshedJwt);
+          fetchToken = refreshedJwt;
+        } else {
+          console.log('🐶 FETCH refreshed token is not valid');
+          logout({ isManual: false });
+          setDataLoading(false);
+          return Promise.reject(undefined);
+        }
+      }
+    }
```
ciaranschutte, line 94: very similar logic here that is already in AuthProvider. 
We can probably keep anything Auth in AuthProvider and just have `userDataContext` worry about the fetching of data

global/utils/auth/fetchEgoJwt.tsx
```diff
@@ -0,0 +1,43 @@
+/*
```
ciaranschutte, line 1: all these small function files in `utils/auth` can just go with auth context mostly. 
they're not really used anywhere else, so you can colocate in a folder ie. `hooks/auth/deletetokens` can go with `hooks/auth/useAuthContext`.  heck you can probably just have one helper file colocated and just do named exports.

global/utils/auth/logoutUtil.tsx
```diff
@@ -20,10 +20,20 @@
 import deleteTokens from './deleteTokens';
```
ciaranschutte, line 20: is `logoutUtil` still in use?
if so the type for `handleTokenState` is fine with just `((jwt: string) => void`

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,160 @@
...
-  const authData = {
-    cancelFetchWithAuth,
-    fetchWithAuth,
-    isLoading,
+  const authContextValue = {
+    handleUserJwt,
     logout,
     permissions,
-    token,
+    setUserLoading,
```
ciaranschutte, line 167: doesn't look like `setUserLoading` this is used anywhere else, no need to export

ciaranschutte (commented):
mostly minor stuff because I think you're working on some bigger changes from the call today.

if you are having trouble with the navbar check the links are using Next link otherwise _app will refresh. also the order of setting state on logout and navigating can be fixed with a useEffect if you're not listening to route events or page context

global/hooks/useAuthContext.tsx
```diff
@@ -17,165 +17,258 @@
...
+    router.events.on('routeChangeStart', handleStart);
+    router.events.on('routeChangeComplete', handleStop);
+    router.events.on('routeChangeError', handleStop);
+
+    return () => {
+      router.events.off('routeChangeStart', handleStart);
+      router.events.off('routeChangeComplete', handleStop);
+      router.events.off('routeChangeError', handleStop);
+    };
+  }, [router]);
```
ciaranschutte, line 234: I don't think you need `forceString`. 
If you look here: https://nextjs.org/docs/api-reference/next/router#routerevents
all these events _except_ route change take url as first param.

`err` object is actually first param for `routeChangeError` so all `forceString` is doing is burying an error.
I think something along the lines of returning empty string which never includes `session_expired=true` so user loading is never set to false

ciaranschutte:
two notes and we can chat more tomorrow.

1)
might be worth dropping in a logger like winston and then you can log to your hearts content and turn if off for prod environments (may even do this by default). removing console.logs while not detrimental is prone to missing some.

2) 
i double checked how page context is working. it'll update on navigation because _app passes the new value to root. so on every root render we get a value. just something to be aware of as I know there's are subsections of pages that you were doing auth calls based on page context.

global/hooks/useAuthContext.tsx
samrichca: Thank you for the insight, I was wondering what's going on!

anncatton (approved):
just a couple little comments, but lgtm, thanks @samrichca !
and thanks @ciaranschutte for your thorough review 😄

anncatton (approved):
Great work @samrichca ! Just a couple small things, plus restoring the Jenkinsfile condition for develop. Also, I'm assuming the logs will be removed prior to tagging to main branch? And we can all test in dev...please correct me if i got that wrong 🙂

PR #572, opened 2022-04-12
Remove terms from validation check

anncatton:
Remove form elements from terms section
Change title to `Introduction` but maintain section name for routing purposes
API will no longer return `terms` as a section so meta will be mocked in form state (WIP). Keep terms in `PRISTINE` state so it does not display any error/locked icons


caravinci:
While I'm fairly certain we can figure out how not to rely on that mock state data, I'm inclined to think this is more than good enough for now. I've tested it in my local, and everything seems to work OK. 👌 

I've opened another PR on top of this one, just to clean up the remaining validation types/seeding data.

PR #574, opened 2022-04-13
Feat/#568 footer fixes

ciaranschutte:


anncatton (approved):
thanks for fixing what i broke

PR #577, opened 2022-04-13
update project info sections to 200 max word

ciaranschutte:


caravinci (approved):
I wanred to be pedantic on the naming consistency, and ask to rename that new variable `laySummarySection...` (i.e. add section), but then realised I just needed to ask you to make fun of me for nitpicking on such silly things instead, @ciaranschutte. 😬

PR #582, opened 2022-04-14
Revert refresh

blabadi:
This PR is to basically reorder commits by revering refresh token changes in develop so they can all go in one PR later, allowing other changes to be in a releasable state for next Monday


anncatton (approved):

blabadi:
Thanks Ann for verifying and testing this locally, @samrichca I'm going to merge now and we can add these commits back when everything is ready

PR #592, opened 2022-04-25
🐛 583 - Persist required field error on all text fields

anncatton:
Fixes bug where the required field error was not persisting on autocomplete fields when focusing/blurring with an empty value
- Adds a property `valueStayedEmpty` to prevent PATCH requests from firing when a field value remains an empty string after the blur event. `shouldPersistResults` was used for this check previously but caused unnecessary api requests on autocomplete fields
- Adds a comment about the different autofill behaviour of FF and Chrome

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
+++ b/components/pages/Applications/ApplicationForm/Forms/validations/index.ts
@@ -465,6 +465,8 @@ export const validator: FormSectionValidatorFunction_Main = (formState, dispatch
         return {
           fieldName,
           results,
+          // shouldPersistResults:true is necessary to get the required field error to persist after other fields are focused/blurred/changed
+          // valueStayedEmpty is used to ensure that fields that remain empty after being touched do not trigger a PATCH request
           shouldPatch: fieldObj.shouldPersistResults && !fieldObj.valueStayedEmpty && !isModalShape,
         };
       });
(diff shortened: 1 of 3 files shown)


caravinci (approved):
LGTM 🚢

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
```diff
@@ -719,6 +717,14 @@ export const useLocalValidation = (
       const isList = sectionName === 'collaborators';
 
       switch (eventType) {
+        /**  **NOTE** Firefox and Chrome handle autofill events differently:
+        Chrome will fire a CHANGE event for each field that is autofilled, this updates local state but does not trigger an api request
+        The api request containing all changed fields will be triggered once the user blurs the focused field
+        Firefox will *immediately* fire a CHANGE and a BLUR event for each field that is autofilled, which triggers an api request *for each* autofilled field
+        github thread here: https://github.com/whatwg/html/issues/3016
+        older SO answer but references the general issue: https://stackoverflow.com/questions/11708092/detecting-browser-autofill/11710295#11710295
+        TODO: add notes about this to the frontend dev wiki
```
samrichca, line 726: I would update the wiki and remove this line

samrichca:
I can't run this locally right now but the logic looks sound and I really appreciate the naming and comments.

samrichca (approved):

components/pages/Applications/ApplicationForm/Forms/validations/index.ts
anncatton: Done-zo

PR #594, opened 2022-04-29
✨ 593 - Maintenance mode with feature flag

anncatton:
Display maintenance page when maintenance mode flag is on.
- `/new-website-notice` route will not display in this mode
- use flag `NEXT_PUBLIC_MAINTENANCE_MODE_ON` in helm


blabadi (approved):

blabadi:
Please merge this to develop and then master before merging the other auto fill bug one ! so each go in different docker images

PR #597, opened 2022-05-09
check for null in collaborators

blabadi:


blabadi:
still haven't tested locally, will merge when tested

anncatton (approved):

PR #601, opened 2022-05-11
583 - Fix empty value check for collabs

anncatton:
Added undefined check for collab value to persist required field error on collab form.
Add optional chaining to the value accessor


samrichca:
Good catch & thanks for leaving a comment!

PR #608, opened 2022-06-07
🐳  add peer-dep flag to dockerfile

anncatton:
so react 17 will play nicely with uikit and jenkins is happy


anncatton:
> @anncatton will probably need to add to jenkins too.

oh yes derp

PR #611, opened 2022-06-09
🐳  Fix npm ci command

anncatton:
latest node 16 version uses npm 8.11.0, which needs the `--legacy-peer-deps` to work with uikit


blabadi (approved):

caravinci (approved):
didn't realise `ci` also accepted that flag. 👍

Issue #486, opened 2021-10-06
[UI] Application when access is paused because attestation hasn’t been completed


kcullion:
## Goal
To transition the application into PAUSED state when the attestation hasn’t been completed:

## Expected Outcomes

Zeplin: https://zpl.io/bloeKXP
- [x] Check the API for application that are in PAUSED state 
- [x] 1. Header change: show Paused date in red
- [x] 2. Header change: show the new progress bar with APPROVED changed to PAUSED
- [x] 3. Lock application for edits
- [x] Keep the banner that was built in https://github.com/icgc-argo/dac-ui/issues/484 

![image](https://user-images.githubusercontent.com/18424138/140531765-66fe69dd-f5bd-4817-bf16-6a84957372ea.png)

b-f-chan:
Released to prod, prod testing not possible for OICR team, hence we will close this ticket, and open new issues if any are found by users in production.

PR #620, opened 2022-07-19
Feat/#484 attestation application state 45 days prior to attestation date

daniel-cy-lu:
In ApplicationForm Header, if application is attestable and hasn't done so(`isAttestable` is TRUE and `attestedAtUtc` is NOT present), 
1. show red pausing date 
2. show pill with text, "Attestation Requested"
3. beneath the Header, show a banner with warning and agreement along with a button with text, "I ATTESTED TO THE ABOVE TERMS"

### DIFF
--- a/components/pages/Applications/ApplicationForm/Forms/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/index.tsx
@@ -199,6 +199,7 @@ const ApplicationFormsBase = ({
                 css={css`
                   margin-top: 20px;
                   margin-left: 10px;
+                  margin-bottom: 20px;
                 `}
               >
                 <span>
@@ -225,6 +226,18 @@ const ApplicationFormsBase = ({
                   4. I will comply with all ethical and regulatory requirements applicable within my
                   institution and country/region in my use of the data.
                 </span>
+                <br />
+                <br />
+                <Button
+                  css={css`
+                    margin-top: 20px;
+                    margin-left: 10px;
+                    margin-bottom: 20px;
+                  `}
+                  size="sm"
+                >
+                  I ATTESTED TO THE ABOVE TERMS
+                </Button>
               </div>
             }
             interactionType="NONE"
--- a/components/pages/Applications/ApplicationForm/Forms/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/index.tsx
@@ -182,12 +182,13 @@ const ApplicationFormsBase = ({
   return (
     <>
       <ContentBody>
-        {requiresAttestation && [ApplicationState.APPROVED].includes(applicationState) && (
+        {/* forward thinking: requiresAttestation is also true for PAUSED state, therefore we want to make sure application is in APPROVED state */}
+        {requiresAttestation && applicationState === ApplicationState.APPROVED && (
           <Notification
             title={
               <div
                 css={css`
-                  margin-top: 5px;
+                  margin-top: 8px;
                   margin-left: 10px;
                 `}
               >
@@ -230,9 +231,8 @@ const ApplicationFormsBase = ({
                 <br />
                 <Button
                   css={css`
-                    margin-top: 20px;
-                    margin-left: 10px;
-                    margin-bottom: 20px;
+                    margin-top: 13px;
+                    margin-bottom: 13px;
                   `}
                   size="sm"
                 >
--- a/components/pages/Applications/ApplicationForm/Forms/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Forms/index.tsx
@@ -86,7 +86,7 @@ const ApplicationFormsBase = ({
   sectionData,
   isAttestable,
   attestedAtUtc,
-  attestationByUtc,
+  attestationByUtc = '',
 }: {
   appId: string;
   applicationState: ApplicationState;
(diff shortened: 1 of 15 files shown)


components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -176,6 +182,70 @@ const ApplicationFormsBase = ({
...
+        {requiresAttestation && applicationState === ApplicationState.APPROVED && (
+          <Notification
+            title={
+              <div
+                css={css`
+                  margin-top: 8px;
+                  margin-left: 10px;
+                `}
+              >
+                Annual Attestation is required by July 28, 2022 or access will be paused
```
caravinci, line 195: Is this date supposed to be fixed/hardcoded for every application?

caravinci (changes requested):
Other than that hardcoded date comment, the changes generally LGTM :+1:
Will defer to @anncatton for feature-specific feedback + approval.

caravinci (approved):

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -176,6 +186,73 @@ const ApplicationFormsBase = ({
...
+          <Notification
+            title={
+              <div
+                css={css`
+                  margin-top: 8px;
+                  margin-left: 10px;
+                `}
+              >
+                {`Annual Attestation is required by ${format(
+                  new Date(attestationByUtc || ''),
```
caravinci, line 200:
```suggestion
                  new Date(attestationByUtc),
```

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -82,6 +84,9 @@ const ApplicationFormsBase = ({
   formState,
   validateSection,
   sectionData,
+  isAttestable,
+  attestedAtUtc,
+  attestationByUtc,
```
caravinci, line 89:
```suggestion
  attestationByUtc = '',
```

caravinci (commented):
Suggestion, not required:
You could make that date line simpler if you make `attestationByUtc` default to empty string when missing.

components/pages/Applications/ApplicationForm/Header/index.tsx
```diff
@@ -59,11 +62,21 @@ const ApplicationHeader = ({
...
     [ApplicationState.REVISIONS_REQUESTED, ApplicationState.SIGN_AND_SUBMIT].includes(state);
 
+  const requiresAttestation = !attestedAtUtc && isAttestable;
+
   // only pass expiry for applications that have been approved
-  const expiry = approvedAtUtc
+  // add 'status' key to allow easy string changes
+  const expiry = requiresAttestation
+    ? {
+        date: format(new Date(attestationByUtc || ''), DATE_TEXT_FORMAT),
```
caravinci, line 71: which could also be done here

components/pages/Applications/ApplicationForm/Forms/index.tsx
daniel-cy-lu: Got it! like what we did when we were coming up with the mock data.

components/pages/Applications/ApplicationForm/Forms/index.tsx
daniel-cy-lu: Good idea. I will make both of your suggested changes.

components/pages/Applications/ApplicationForm/Header/index.tsx
daniel-cy-lu: I will make this change as well!

components/pages/Applications/ApplicationForm/Header/index.tsx
```diff
@@ -59,11 +62,21 @@ const ApplicationHeader = ({
...
+  const expiry = requiresAttestation
+    ? {
+        date: format(new Date(attestationByUtc), DATE_TEXT_FORMAT),
+        isExpired: true,
+        status: '! Pausing',
+      }
+    : approvedAtUtc
     ? {
-        date: format(new Date(closedAtUtc || expiresAtUtc || ''), DATE_TEXT_FORMAT),
+        date: format(new Date(closedAtUtc || expiresAtUtc), DATE_TEXT_FORMAT),
```
caravinci, line 77: +Going above and beyond! 💫

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -176,6 +186,73 @@ const ApplicationFormsBase = ({
   return (
     <>
       <ContentBody>
+        {/* forward thinking: requiresAttestation is also true for PAUSED state, therefore we want to make sure application is in APPROVED state */}
```
anncatton, line 189: i think the banner is the same when the app is in paused state so we may not need this check...however i can check with @kcullion 🙂 
but still a good thought, and thanks for the helpful comment!

anncatton (approved):
Great work @daniel-cy-lu ! We can just check that one comment I made about paused state that's going to come up in #486

components/pages/Applications/ApplicationForm/Forms/index.tsx
kcullion: yes the banner remains on the application when the access is paused...and the application is locked. mockup: https://zpl.io/bloeKXP

daniel-cy-lu: The same banner is used in '45 days period' and in 'Paused' state. The warning banner now appears regardless of the state. It appears whenever the application needs attestation to be done.

PR #622, opened 2022-07-20
add 'PAUSED' state to variable adminStatesAllowList to allow PAUSED s…

daniel-cy-lu:
add 'PAUSED' state to variable adminStatesAllowList to allow PAUSED state to show in 'Status' column


anncatton (approved):
👍

Issue #484, opened 2021-10-06
[UI] Attestation application state 45 days prior to attestation date


kcullion:
## Goal 
Build the UI for when the applicant needs to attest at the 1 year mark.

**Related API ticket**: https://github.com/icgc-argo/dac-api/issues/236
https://github.com/icgc-argo/dac-api/issues/240

## To do
Zeplin: https://zpl.io/2jdyAe6
- [x] Applications will return a `attestedAtUtc` value if the user has attested. If it's not present, assume attestation needs to be done if `isAttestable` is TRUE
- [x] 1. If so, show pill/flag (this is the same as request revisions pill - have a look at that)
- [x] 2. Show pausing date 
- [x] 3. Show Banner above application only on the applicants' side and not on the admin side (Use UI kit for this banner. We have implemented some success banners if you want to have a look at those)
Note: application should still be in approved state - so locked with some sections editable
![image](https://user-images.githubusercontent.com/18424138/140531581-974e1a9a-fcd4-4edf-937b-37db34aecf8d.png)

b-f-chan:
Released to prod, prod testing not possible for OICR team, hence we will close this ticket, and open new issues if any are found by users in production.

PR #624, opened 2022-07-29
feat/#484-display-attestation-banner-for-applicant-only

daniel-cy-lu:
- continuing from the merged #484 branch
- using isAdmin prop to prevent admin side from seeing the attestation banner


anncatton (approved):
🎉

PR #625, opened 2022-07-29
create functionality for attestation button inside the banner

daniel-cy-lu:


components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -185,6 +193,26 @@ const ApplicationFormsBase = ({
     );
   };
 
+  const { fetchWithAuth } = useAuthContext();
+  const handleSubmitAttestion = () => {
+    fetchWithAuth({
+      data: {
+        //This works for testing. Assuming API handles the New Date object
```
daniel-cy-lu, line 200: @anncatton Can you double check the data (attestedAtUtc) I am sending to the API. Right now I am sending a new Date object back to the API but I think I am supposed to send a string.

caravinci (approved):
LGTM :shipit: 
(Did not test it.)

PR #629, opened 2022-09-26
♻️✏️ 484 - Refactoring for app access info, typo fix

anncatton:
Minor fixes/refactoring for `PAUSED` app ui
- fixes typo
- adds `PAUSED` app case for PDF button text
- refactors access info comp to account for other conditions besides expiry
- moaarr types
- uses `isAttestable` value from api instead of doing any assessment from ui side


caravinci (approved):
LGTM 👍

PR #634, opened 2022-10-06
🩹 486 - correct action buttons for paused state

anncatton:
Shows "Close Application" button in paused state
Shows PDF action buttons to admin in paused state


caravinci (approved):
👍

PR #642, opened 2023-02-06
💬 status text update

anncatton:
Just clarifying the language for the status text on a pausing application


b-f-chan (approved):
Lgtm!  Thanks for modifying

PR #646, opened 2023-02-20
✨ 504 - Add expiring status, renew button to dashboard card

anncatton:
Dashboard status text, date and renew button for approved applications within the renewal period (before expiry).
Refactors some of the conditionals for status display.
Allows a custom Button component
Adds calc function for renewal period end date (expiry + 90 days) - currently this is hardcoded, either switch to a configurable value, or get this value from the BE in some form (a new field or retrieve config from api)

### DIFF
--- a/components/pages/Applications/Dashboard/Applications/InProgress/ButtonGroup.tsx
+++ b/components/pages/Applications/Dashboard/Applications/InProgress/ButtonGroup.tsx
@@ -17,15 +17,20 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { Fragment } from 'react';
+import { Fragment, ReactNode } from 'react';
 import { css } from '@emotion/core';
 import Button from '@icgc-argo/uikit/Button';
 import Icon from '@icgc-argo/uikit/Icon';
 import { ApplicationState } from 'components/ApplicationProgressBar/types';
 import router from 'next/router';
-import { APPLICATIONS_PATH } from 'global/constants';
+import { API, APPLICATIONS_PATH, RENEWAL_PATH } from 'global/constants';
 import urlJoin from 'url-join';
 import { UikitIconNames } from '@icgc-argo/uikit/Icon/icons';
+import { useAuthContext } from 'global/hooks';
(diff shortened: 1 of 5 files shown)


caravinci:
Quick question: what's the difference between `appId` and `renewalAppId`? is the latter an id for a different type of app than the "regular" one?

components/pages/Applications/Dashboard/Applications/InProgress/index.tsx
```diff
@@ -43,8 +43,78 @@ export interface StatusDates {
   attestedAtUtc: string;
   attestationByUtc: string;
   lastPausedAtUtc?: string;
+  expiresAtUtc: string;
 }
 
+const getStatusDate = (application: ApplicationSummary): any => {
```
caravinci, line 49: 👍 three cheers for well implemented abstraction!

caravinci:
Oh so there is in fact something called a "Renewal app"? awesome. Thanks for clarifying!
DACO has reaaally changed since I was involved.

components/pages/Applications/Dashboard/Applications/InProgress/ButtonGroup.tsx
```diff
@@ -62,18 +124,17 @@ const getButtonConfig = (
         {
           content: 'Edit Application',
           link,
-          icon: icons.edit,
+          icon: 'edit',
```
caravinci, line 127: 🤔 wonder whether it'd be useful/desirable to use enum properties here, rather than string literals.

caravinci (approved):
LGTM :shipit:

anncatton:
haha yeah @caravinci renewals were in the initial roadmap but this implementation is much different than what we had planned 🙂

components/pages/Applications/Dashboard/Applications/InProgress/ButtonGroup.tsx
anncatton: these are from uikit, tho? which maybe has changed its implementation recently, i'm not sure - dac-ui will need to update that dep once these last features are completed

caravinci: ah fair enough. It likely would be useful for uikit to export said Enum, then.
Will keep this in mind later for a new component library we're starting to plan for.
Thanks, @anncatton!

PR #649, opened 2023-03-04
💄 506 - renewal app card

anncatton:
Adds custom text/status for renewal cards in `DRAFT`, `SIGN AND SUBMIT` or `REVISIONS REQUESTED`
Adds status date as well, but need to decide what the display text should be.


headsoup (approved):
The business logic intensifies... 🧐

PR #650, opened 2023-03-04
renew button hookup

anncatton:
Adds Renew button to application header when app is `ableToRenew`
- adds banners related to renewal status for both renewal and source applications. Includes some `TODO`s for some outstanding questions, will be reviewing these with @b-f-chan and @patrickdsantos 
- some minor refactoring
- adds `isWithinRenewalPeriod` helper func
- text change for submitting a renewal
- may need some improvements/feedback on loading transition when triggering from single application view (see comment below)


components/pages/Applications/ApplicationForm/index.tsx
```diff
@@ -62,19 +62,20 @@ const ApplicationForm = ({ appId = 'none', isAdmin = false }): ReactElement => {
     <Loader />
   ) : data ? (
     <>
-      <ApplicationHeader refetchAllData={formState.__refetchAllData} data={data} />
+      <ApplicationHeader
+        refetchAllData={formState.__refetchAllData}
+        data={data}
+        setIsAppLoading={setIsAppLoading}
```
anncatton, line 68: Loading states are not working, these props should prob be removed and loading for a renewal handled further up

components/pages/Applications/index.tsx
```diff
@@ -54,7 +54,7 @@ const Application = (): ReactElement => {
       {isLoading ? (
         <Loader />
       ) : appId ? (
-        <ApplicationForm appId={normalisedAppId} isAdmin={isAdmin} />
+        <ApplicationForm key={appId} appId={normalisedAppId} isAdmin={isAdmin} />
```
anncatton, line 57: Clicking the "Renew" button in the single application view updates the url `ID` param, but was not updating the ApplicationForm content. Adding the `key` prop forces the ApplicationForm component to remount when the `appId` changes, and the form data/header info/validations get updated correctly. AFAIK this is the only scenario in the UI where there's a direct appId switch without a hard page reload. The loader here *should* trigger but has been difficult to see locally, it may be easier to see when testing in the dev env; this may need some finessing so the transition is more obvious to the user.

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -81,36 +86,192 @@ const getActiveSection = (sectionFromQuery?: FormSectionNames): FormSectionNames
...
+              DateFormat.DATE_TEXT_FORMAT,
+            )}.`}
+            css={notificationStyle}
+          />
+        );
+      }
+      break;
+
+    case state === ApplicationState.EXPIRED && isRenewalPeriodEnded(expiresAtUtc):
+      if (!renewalAppId) {
```
headsoup, line 203: wouldn't it not be harder to read this condition if it wasn't not negative? 🤣

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -225,7 +386,7 @@ const ApplicationFormsBase = ({
                 `}
               >
                 {`Annual Attestation is required by ${getFormattedDate(
-                  attestationByUtc,
+                  attestationByUtc as string,
```
headsoup, line 389: curious about what makes this casting necessary here.
does this imply we could change `attestationByUtc` in `ApplicationData`?

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -270,7 +431,7 @@ const ApplicationFormsBase = ({
                     margin-bottom: 13px;
                   `}
                   size="sm"
-                  onClick={handleSubmitAttestion}
+                  onClick={handleSubmitAttestation}
```
headsoup, line 434: nice catch! 💪

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -351,6 +515,7 @@ const ApplicationFormsBase = ({
             css={notificationStyle}
           />
         )}
+
```
headsoup, line 518: 👍

components/pages/Applications/index.tsx
headsoup: nice workaround. easy change with good benefits!

headsoup (approved):

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -81,36 +86,192 @@ const getActiveSection = (sectionFromQuery?: FormSectionNames): FormSectionNames
...
+    default:
+      return null;
+  }
+};
+
 const ApplicationFormsBase = ({
   appId = 'none',
-  applicationState,
+  appData,
```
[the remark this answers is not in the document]
anncatton: i remember that conversation lol

components/pages/Applications/ApplicationForm/Forms/index.tsx
anncatton: yes i feel like *that* discussion would fill up a whole series of podcasts 😛

components/pages/Applications/ApplicationForm/Forms/index.tsx
anncatton: actually this shouldn't need to be cast, i had made changes with the date formatter for this reason. `attestationByUtc` is correctly typed, it is possibly undefined. Nice catch, I'll fix this!

components/pages/Applications/ApplicationForm/Forms/index.tsx
anncatton: oh man that's the perfect snobby-sounding title 😂

components/pages/Applications/ApplicationForm/Forms/index.tsx
anncatton: anyone who's worked on the attestation feature has spelled/pronounced it incorrectly at least 5 times

headsoup: oh I'd probably have made it the sixth!

components/pages/Applications/ApplicationForm/Forms/index.tsx
```diff
@@ -331,7 +330,9 @@ const ApplicationFormsBase = ({
   const sectionsAfter = enabledSections(sectionsOrder.slice(sectionIndex + 1), formState);
   const sectionsBefore = enabledSections(sectionsOrder.slice(0, sectionIndex), formState);
 
-  const requiresAttestation = !attestedAtUtc && isAttestable;
+  // isAttestable being true implies that an attestationByUtc date is present
+  // but adding attestationByUtc here to make TS happy in date formatter call below
+  const requiresAttestation = isAttestable && attestationByUtc;
```
headsoup, line 335: 👍 this not only makes TS happy, but also makes the attestation logic in this spot more understandable.
nice comment!

anncatton: although the ts build step is not happy! i guess it's more strict, will fix

PR #658, opened 2023-03-11
♻️ Refactor date formatters

anncatton:
Reorganize date format constants into enums, add to global/dates folder.
Move `getFormattedDate` helper func to global/dates folder. Add error clause.
Replace most `date-fns/format` calls with helper func.
**Does not modify/remove DateFormat string options, should discuss whether any of these can be pruned first.**

### DIFF
--- a/components/pages/Applications/ApplicationForm/Header/index.tsx
+++ b/components/pages/Applications/ApplicationForm/Header/index.tsx
@@ -22,7 +22,7 @@ import { format } from 'date-fns';
 import { css } from '@icgc-argo/uikit';
 import { UikitTheme } from '@icgc-argo/uikit/index';
 import PageHeader from 'components/PageHeader';
-import { DATE_TEXT_FORMAT } from 'global/utils/dates/constants';
+import { DateFormat } from 'global/utils/dates/types';
 import Actions from './Actions';
 import Details from './Details';
 import Progress from './Progress';
@@ -30,6 +30,7 @@ import { RefetchDataFunction } from '../Forms/types';
 import { ApplicationState } from 'components/ApplicationProgressBar/types';
 import { ApplicationData } from '../../types';
 import { isPastExpiry } from '../Forms/helpers';
+import { getFormattedDate } from 'global/utils/dates/helpers';
 
 export type ApplicationAccessInfo = { date?: string; isWarning: boolean; status: string };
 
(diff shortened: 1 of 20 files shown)


components/pages/Applications/ApplicationForm/Header/index.tsx
```diff
@@ -133,8 +135,11 @@ const ApplicationHeader = ({
         <Details
           appId={appId}
           applicant={applicant}
-          createdAt={format(new Date(createdAtUtc), DATE_TEXT_FORMAT)}
-          lastUpdated={format(new Date(lastUpdatedAtUtc), DATE_TEXT_FORMAT + ' h:mm aaaa')}
+          createdAt={getFormattedDate(createdAtUtc, DateFormat.DATE_TEXT_FORMAT)}
+          lastUpdated={format(
+            new Date(lastUpdatedAtUtc),
```
headsoup, line 140: out of sheer curiosity, why this one without `getFormattedDate`?

global/utils/dates/types.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 The Ontario Institute for Cancer Research. All rights reserved
+ * Copyright (c) 2023 The Ontario Institute for Cancer Research. All rights reserved
```
headsoup, line 2: lol I was joking earlier on how @ciaranschutte did the same year update in a single file in a PR, and that it was a clue to an easter egg. Are you acting together in this mystery?
Is that the case here too? am I getting closer to the egg?
so many questions! 😆

headsoup (approved):
much standard. such unified! 🐶

components/pages/Applications/ApplicationForm/Header/index.tsx
anncatton: this one has that additional ` h:mm aaaa` appended, which i didn't add to the `DateFormat` type, so `getFormattedDate` would yell...I was thinking this might be one of the formats that can get dropped.

anncatton: In fact I should really leave a comment about that 😛

global/utils/dates/types.ts
anncatton: 🥚 🥚 🥚

PR #660, opened 2023-03-20
💄 515 - admin table - move type column after app status column

anncatton:
Move type column after app status column in admin table view so status shows without need for horizontal scroll.


headsoup (approved):
😓 so change

PR #673, opened 2023-04-13
✨ 665 - add confirm action when closing an app

anncatton:
Adds extra confirm text input when closing an application to prevent accidental close.


components/pages/Applications/ApplicationForm/Header/Actions.tsx
```diff
@@ -163,32 +170,41 @@ const HeaderActions = ({
...
+                )}
 
                 <Button
                   variant="text"
                   css={css`
                     margin-left: 10px;
                   `}
                   size="md"
-                  onClick={dismissModal}
+                  onClick={(e) => {
```
justincorrigible, line 203: may I suggest abstracting this function from the return/render, rather than defining it inline? that way a new function won't be defined everytime there's a re-render

justincorrigible (approved):

justincorrigible (approved):

PR #674, opened 2023-04-13
💄 590 - More padding for project info list items

anncatton:
More padding for project info static content, fixes text running outside border

### DIFF
--- a/components/pages/Applications/PDF/common.tsx
+++ b/components/pages/Applications/PDF/common.tsx
@@ -99,7 +99,7 @@ export const styles = StyleSheet.create({
   },
   bubbleContainer: {
     border: `1px solid ${defaultTheme.colors.grey_1}`,
-    padding: '10pt 5pt',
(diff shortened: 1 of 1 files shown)


components/pages/Applications/PDF/common.tsx
```diff
@@ -99,7 +99,7 @@ export const styles = StyleSheet.create({
   },
   bubbleContainer: {
     border: `1px solid ${defaultTheme.colors.grey_1}`,
-    padding: '10pt 5pt',
+    padding: '10pt',
```
justincorrigible, line 102: 😓 please, when there's a big change like this, let's break the PR into smaller chunks so it's easy to review! 😉😄

justincorrigible (approved):

components/pages/Applications/PDF/common.tsx
anncatton: 😂 😂

PR #677, opened 2023-04-18
💄 675 - Ensure proper margin for dashboard renew button

anncatton:


components/pages/Applications/Dashboard/Applications/InProgress/ButtonGroup.tsx
```diff
@@ -62,6 +64,9 @@ export const RenewButton = ({
     <Button
       className="action-btns"
       size="sm"
+      css={css`
+        margin-left: ${index ? '8px !important;' : 0};
```
justincorrigible, line 68: while I'm not 100% sure I'm reading the intent correctly, looks like this could better be achieved with a css selector:

`&:not(:first-of-type) { ... }`

anncatton: much better! i can get rid of the `important` too

justincorrigible (approved):

